PR #1, opened 2020-01-13
Rename to NervesHubLinK

jjcarstens:
The short history of this is that `nerves_hub` library had a few shortcomings we wanted to overcome such as the slight confusion in the naming and the intermixing of `http` and `channel` support making unique edge-cases.

An experiment was done which forked `nerves_hub` into `nerves_hub_device` to test a few notable ideas:
  * Remove `HTTP` support to use channels exclusively
  * Reorganize as a an `Application` and remove ability to start a supervisor separately
  * Support using `NervesKey` via a configuartion flag (for ease of use)

We've been using this at Smartrent and after the last [Nerves stand-up meeting](https://github.com/nerves-project/nerves/wiki/Core-Team-Stand-Up-Meeting-Notes#jon-carstens) decided rename it to `NervesHubLink` and bring it over to NervesHub scope, which is what this PR does.

I've kept the fork up-to-date with changes in `NervesHub` that pertain to channels, so we shouldn't be missing any existing features.

### DIFF
--- a/.iex.exs
+++ b/.iex.exs
@@ -1,1 +1,1 @@
-alias NervesHubDevice.FirmwareChannel
+alias NervesHubLink.FirmwareChannel
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # NervesHub
 
-[![CircleCI](https://circleci.com/gh/nerves-hub/nerves_hub_device/tree/master.svg?style=svg)](https://circleci.com/gh/nerves-hub/nerves_hub_device/tree/master)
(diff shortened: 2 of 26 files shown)


test/test_helper.exs
```diff
@@ -1 +1 @@
-ExUnit.start()
+ExUnit.start(capture_log: true)
```
fhunleth: This change seems unrelated to the name change.

jjcarstens: Derp. Forgot that was in there. This was me not wanting to see all the log messages when running the tests. I'll break it out

jjcarstens: Fixed

PR #3, opened 2020-02-06
Add ability to make runtime configuration changes for the socket

jjcarstens:
Based on an offline discussion with @mobileoverlord - `NervesKey` would bring in i2c deps which actually breaks things in some x86_64 setup.

So, this abstracts that NervesKey niceness setup into its module along with a default configurator. It also allows the user to tie into the configuration step before attempting to connect to the socket in order to do things required at runtime and not in a static config.

**Edit:** Talked with Justin a bit. We decided to go with a `ssl:` key and compile that into the socket_opts. Makes things a lot cleaner:

```elixir
%Config{
  ssl: [cert: cert, key: key, cacerts: cacerts]
}
```

~Right now, it wants a `:socket` key in the configuration to pass on to the `phoenix_client` socket to support all the options available there. As a result, the way to get your certs at runtime would end up looking like something like this:~

~Not sure how much I love that but allows us to be less restrictive on getting the socket what it needs. Would love feedback on it.~

Also, test dir was still named `nerves_hub_device` so renamed that in 89affa8


jjcarstens:
Now I'm actually thinking maybe the config key should be `:ssl` and `nerves_hub_link` will just be in charge of compiling the opts for `phoenix_client` socket instead of exposing that.. 🤔 thoughts?

lib/nerves_hub_link/configurators/nerves_key.ex
```diff
@@ -0,0 +1,54 @@
+defmodule NervesHubLink.Configurator.NervesKey do
```
mobileoverlord: We should wrap the defmodule here in `Code.ensure_loaded?(:nerves_key)` to prevent the module from being compiled with bad x_refs if `nerves_key` is not loaded.

jjcarstens: Instead of ignoring bad xrefs in `mix.exs`?

jjcarstens: Done

PR #12, opened 2020-04-08
Include support for connecting to staging env

mobileoverlord:
This also moves the certificates from the priv dir to the ssl directory. The CA Certificates are compiled into the module for runtime. This will prevent them fomr being included multiple times for multiple locations.


fhunleth:
I feel like we've been burned a few times by using the application environment at compile time in dependencies.

Can the certificates be included in the default application environment in the `mix.exs` and read at runtime? If someone wants to replace them, they set them in the config.exs like now? Or is there some other constraint that you're dealing with?

mobileoverlord:
The choice to compile in the default set of certificates was made because of confusion around compile time usage / runtime usage of the Mix config.exs. There are a couple things we were trying to accomplish:

  * Allow the ca certificates to be overridden
  * Make is easy for nerves-hub developers to switch between environments (dev/staging/prod)

We could switch to where the path that is listed in the application config has to be a reference to the location on the target. In that case we need to document that references to `priv_dir` locations / locations on host are no longer used. Then we could fall back to the priv dir. NervesHub developers would then need to use `rootfs_overlay` to switch between environments.

fhunleth:
How about not use paths at all and have the cert contents in the application config? I agree that using paths is a source of confusion. 

I think you can accomplish your list of items without compile-time configuration and specifying a default application environment in the `mix.exs`. I'll ping you and show you what I'm talking about...

mobileoverlord:
I'd like to revisit this when we add docs and information for how to spin up a private NervesHub instance. Currently this feature is only used by nerves-hub developers.

PR #17, opened 2020-04-15
Abstract the update code from the device_channel.

mobileoverlord:
This is a start to detangling the update code from the device channel. The next step is to separate the fwup process from the http process.

### DIFF
--- a/lib/nerves_hub_link/client.ex
+++ b/lib/nerves_hub_link/client.ex
@@ -81,9 +81,7 @@ defmodule NervesHubLink.Client do
   """
   @spec update_available(update_data()) :: update_response()
   def update_available(data) do
-    ret = apply_wrap(mod(), :update_available, [data])
-
-    case ret do
+    case apply_wrap(mod(), :update_available, [data]) do
       :apply ->
         :apply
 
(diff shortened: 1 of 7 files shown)


lib/nerves_hub_link/client.ex
```diff
@@ -81,7 +81,9 @@ defmodule NervesHubLink.Client do
   """
   @spec update_available(update_data()) :: update_response()
   def update_available(data) do
-    case apply_wrap(mod(), :update_available, [data]) do
+    ret = apply_wrap(mod(), :update_available, [data])
+
+    case ret do
```
jjcarstens, line 86:
```suggestion
    case apply_wrap(mod(), :update_available, [data]) do
```

Just because I know this was done while troubleshooting 🤣

lib/nerves_hub_link/device_channel.ex
```diff
@@ -69,8 +84,8 @@ defmodule NervesHubLink.DeviceChannel do
     case Channel.join(socket, topic, params) do
       {:ok, reply, channel} ->
         NervesHubLink.Connection.connected()
-        state = %{state | channel: channel, connected?: true}
-        {:noreply, maybe_update_firmware(reply, state)}
+        _ = handle_join_reply(reply)
```
jjcarstens, line 87: This is minor, but I think we could skip defining this new function here and use `UpdateManager.apply_update(reply)` directly since it utilizes the same logic.

Unless this is predatory to changing NervesHub join reply?

PR #22, opened 2020-05-14
provisioning.conf is unused so remove it

fhunleth:
This also fixes a build error when the path couldn't be expanded since
it was wrong.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -1,12 +1,6 @@
 defmodule NervesHubLink.MixProject do
   use Mix.Project
 
-  Application.put_env(
-    :nerves_hub_link,
-    :nerves_provisioning,
-    Path.expand("priv/provisioning.conf")
-  )
-
   def project do
     [
       app: :nerves_hub_link,
@@ -69,8 +63,7 @@ defmodule NervesHubLink.MixProject do
         "CHANGELOG.md",
         "LICENSE",
         "mix.exs",
-        "README.md",
-        "provisioning.conf"
+        "README.md"
       ]
     ]
   end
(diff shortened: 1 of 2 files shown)


mobileoverlord:
provisioning.conf is unused by this app, but it could be used by other application when using NervesHubCA generated certificate files.

mobileoverlord:
See: https://github.com/nerves-project/nerves/search?q=provisioning&unscoped_q=provisioning

fhunleth:
Oh...  I think we should remove those too. I really think this is a bad idea.

fhunleth:
I suppose that I should be more explicit with my critique. I think two things are bad ideas: 1. apps reaching into other application's environment and 2. libraries providing provisioning files. The first is probably a personal preference of mine. The second is problematic since as it is set up, there can only be one `provisioning.conf`. If a library uses it, then project's can't. It also feels like you want provisioning information to be explicit so hiding it in a library feels dangerous. Plus, it's not like you can upgrade provisioning information. Once a device is programmed the first time, this is locked in forever (i.e., firmware updates can't change it unless you do something custom).

mobileoverlord:
We would have to clean up references in the docs too: https://docs.nerves-hub.org/nerves-hub/setup/adding-nerveshub-to-your-project

fhunleth:
Ahh. Shoot. Sigh. I'd love to rage delete this so that we could make it harder for people to store their private keys in the clear.

However, I need to unbreak the build. Let me submit an alternative PR that does the minimum.

fhunleth:
See #23

PR #24, opened 2020-05-14
Fix calling undefied default client module

jjcarstens:
I think this was initially thought to be aliased, but it wasnt so this was throwing a `{:error, :undef}` error when trying to call `update_available/1` on the deduced default client module


fhunleth:
The 'n' key on your keyboard is having trouble.

PR #35, opened 2020-07-07
Change ConsoleChannel terminal emulation

jjcarstens:
This completely changes the remote console API to support a more complete terminal emulation via `ExTTY`. This allows us to more easily emulate features like tab completion and history search in the browser.

This will need a NervesHub deploy before it can be released and used.


ConnorRigby:
:shipit:

lib/nerves_hub_link/console_channel.ex
```diff
@@ -3,73 +3,63 @@ defmodule NervesHubLink.ConsoleChannel do
...
               channel: nil,
               params: [],
-              iex_pid: nil,
-              request: nil,
-              retry_count: 0
+              iex_pid: nil
   end
 
-  def start_link(opts), do: GenServer.start_link(__MODULE__, opts)
+  def start_link(opts), do: GenServer.start_link(__MODULE__, opts, name: __MODULE__)
```
fhunleth, line 48: Not that it really matters, but what's causing this to become a singleton?

jjcarstens: O 🤦 

This is me doing `:sys.get_state(NervesHubLink.ConsoleChannel)` during troubleshooting/development. It doesn't need to be a singleton, its just convenient for my lazy self

lib/nerves_hub_link/console_channel.ex
```diff
@@ -80,17 +70,19 @@ defmodule NervesHubLink.ConsoleChannel do
...
+    :ok = GenServer.stop(state.iex_pid, 10_000)
+    {:ok, iex_pid} = ExTTY.start_link(handler: self(), type: :elixir)
+    {:noreply, %{state | iex_pid: iex_pid}}
+  end
 
-  def handle_info(%Message{event: "io_reply", payload: %{"data" => data, "kind" => kind}}, state) do
-    {from, reply_as, _} = state.request
-    data = if kind == "put_chars", do: String.to_existing_atom(data), else: data
-    io_reply(from, reply_as, data)
+  def handle_info(%Message{event: "tty_text", payload: %{"data" => data}}, state) do
```
fhunleth, line 84: It looks like the message is called `tty_data` for data that goes to the server and `tty_text` for data that comes back. How about a short name? Maybe the same word for both directions since `tx` and `rx` might be confusing? Maybe `io`?

Does the payload need to be a map or could it be a string?

In addition to simplicity, I'd like to minimize the starting overhead of using the console. I feel like other optimizations will help out way more, like implementing Nagle's algorithm, but that's more work, and reducing naming here seems easy before the API is set.

jjcarstens: I tried with just a binary payload and got this error
```elixir
[error] GenServer #PID<0.1282.0> terminating
** (ArgumentError) topic and event must be strings, message must be a map, got:

  topic: "console:10"
  event: "tty_data"
  payload: "Interactive Elixir (1.10.3) - press Ctrl+C to exit (type h() ENTER for help)\r\n"


    (phoenix 1.4.16) lib/phoenix/channel/server.ex:200: Phoenix.Channel.Server.raise_invalid_message/3
    (nerves_hub_device 0.1.0) lib/nerves_hub_device_web/channels/console_channel.ex:41: NervesHubDeviceWeb.ConsoleChannel.handle_in/3
    (phoenix 1.4.16) lib/phoenix/channel/server.ex:280: Phoenix.Channel.Server.handle_info/2
    (stdlib 3.13) gen_server.erl:680: :gen_server.try_dispatch/4
    (stdlib 3.13) gen_server.erl:756: :gen_server.handle_msg/6
    (stdlib 3.13) proc_lib.erl:226: :proc_lib.init_p_do_apply/3 
```

It looks like the channel function `handle_in/3` requires the payload to be a map. However, we might be able to drop in our own serializer to skip the map requirement? It would take a little bit of investigation.

I'll have to think about the event name being the same as `io` - It seems like that might cause some issues since we're doing a lot of broadcasting around nodes and back to the device, but I just need to feel it out a bit

jjcarstens:
Changed the message names to match `nerves_hub_web`:

* `tty_data` => `up`
* `tty_text` => `dn`

This should be ready for another look

mobileoverlord:
I am deploying the nerves_hub_web PR to staging so I can test this out. Taking a peek 👀

mobileoverlord:
🎉

Issue #43, opened 2020-08-28
Report FWUP version to NervesHub


Mrjaco12:
In [certain scenarios](https://github.com/nerves-hub/nerves_hub_web/issues/538) it makes sense for NervesHub to know what version of FWUP a device is running this means FWUP needs to report this data in it's communications with the NervesHub back end.

Mrjaco12:
@fhunleth @mobileoverlord I'm creating this issue to track adding the FWUP version here and, I assume, we'll also need it in [the http version](https://github.com/nerves-hub/nerves_hub_link_http). I figured this would be a good place to discuss the approach since you guys have much greater familiarity with the whole Nerves/NervesHub ecosystem than I do.

After a first inspection, my first idea was to just grab the version from fwup via `System.cmd` and insert it around [the time we grab other meta data from the runtime KV](https://github.com/nerves-hub/nerves_hub_link/blob/426dcd64a32af4a3e658ba6a1824cd3bdd865f9c/lib/nerves_hub_link/configurator.ex#L51). Does that make sense or is there some point, before or after, that would work better?

fhunleth:
I like your idea.

jjcarstens:
Released in https://github.com/nerves-hub/nerves_hub_link/releases/tag/v0.9.1

PR #44, opened 2020-08-31
Report fwup version to nerves hub

Mrjaco12:
Why:

* NH needs to know the fwup version to make decisions about delta updates
* Resolves Issue #43

This change addresses the need by:

* Grab fwup version when creating config
* Add fwup version to config params
* Add a spec to check that fwup version key is present

### DIFF
--- a/lib/nerves_hub_link/configurator.ex
+++ b/lib/nerves_hub_link/configurator.ex
@@ -48,7 +48,7 @@ defmodule NervesHubLink.Configurator do
       |> Keyword.put_new(:verify, :verify_peer)
       |> Keyword.put_new(:server_name_indication, to_charlist(base.device_api_sni))
 
-    params = Map.put(Nerves.Runtime.KV.get_all_active(), "nerves_fw_fwup_version", fwup_version())
+    params = Map.put(Nerves.Runtime.KV.get_all_active(), "fwup_version", fwup_version())
 
     %{base | params: params, socket: socket, ssl: ssl}
   end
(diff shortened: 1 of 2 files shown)


Mrjaco12:
@mobileoverlord @fhunleth let me know what you guys think about this. The test is pretty naive and would fail in a testing env without fwup installed so that's a downside but the tradeoff of extracting a module and creating a mock for this didn't really seem worth it.

fhunleth:
IMHO, fwup should be installed everywhere, so I think we should just view that as a given. ;)

lib/nerves_hub_link/configurator.ex
```diff
@@ -48,7 +48,9 @@ defmodule NervesHubLink.Configurator do
       |> Keyword.put_new(:verify, :verify_peer)
       |> Keyword.put_new(:server_name_indication, to_charlist(base.device_api_sni))
 
-    %{base | params: Nerves.Runtime.KV.get_all_active(), socket: socket, ssl: ssl}
+    params = Map.put(Nerves.Runtime.KV.get_all_active(), "nerves_fw_fwup_version", fwup_version())
```
fhunleth, line 51:
```suggestion
    params = Map.put(Nerves.Runtime.KV.get_all_active(), "fwup_version", fwup_version())
```

I'm not sure that the `nerves_fw_` prefix matters here. My preference would be to drop it, but I don't feel strongly about it.

lib/nerves_hub_link/configurator.ex
```diff
@@ -71,4 +73,9 @@ defmodule NervesHubLink.Configurator do
       Default
     end
   end
+
+  defp fwup_version do
+    {version_string, 0} = System.cmd("fwup", ["--version"])
+    version_string |> String.trim()
```
jjcarstens, line 79:
```suggestion
String.trim(version_string)
```
piping for a short call might be unnecessary

PR #50, opened 2020-11-21
v0.9.2 release

jjcarstens:


.circleci/config.yml
```diff
@@ -35,6 +35,37 @@ defaults: &defaults
   working_directory: ~/repo
 
 jobs:
+  build_elixir_1_11_otp_23:
+    docker:
+      - image: erlang:23.1.3
+        environment:
+          ELIXIR_VERSION: 1.11.2-otp-23
```
fhunleth, line 42: Caching and dialyzer need to be removed for the elixir 1.10 build or there's a race on the ci cache.

fhunleth (changes requested):

.circleci/config.yml
fhunleth: Fixed in #51

fhunleth (approved):

PR #53, opened 2020-12-03
add NervesHubLink.reconnect/0

jjcarstens:
For the cases you might want to bump the socket and restart connections to NH (such as network interface changing, etc)


ConnorRigby (approved):

jjcarstens:
Closing this for now as Slipstream has reconnect mechanisms and the goal is to use that for channel connections in #68.

Once implemented, we can revist the reconnect case in a separate PR

Issue #52, opened 2020-12-03
NervesHubLink Fails to Start


psteininger:
I was trying to get a certificate set up on the ATECC608A today, and I ran into the following error:
```
15:33:26.670 [info]  Application nerves_hub_link exited: exited in: NervesHubLink.Application.start(:normal, [])
    ** (EXIT) exited in: GenServer.call(ATECC508A.Transport.I2CSupervisor, {:start_child, {{ATECC508A.Transport.I2CServer, :start_link, [["i2c-1", 96, :"Elixir.ATECC508A.Transport.I2C.i2c-1.96"]]}, :permanent, 5000, :worker, [ATECC508A.Transport.I2CServer]}}, :infinity)
        ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
```
I also get this when I run this from ssh console on the device:
```
iex(2)> Application.ensure_all_started(:thirsty)
{:error,
 {:nerves_hub_link,
  {:bad_return,
   {{NervesHubLink.Application, :start, [:normal, []]},
    {:EXIT,
     {:noproc,
      {GenServer, :call,
       [
         ATECC508A.Transport.I2CSupervisor,
         {:start_child,
          {{ATECC508A.Transport.I2CServer, :start_link,
            [["i2c-1", 96, :"Elixir.ATECC508A.Transport.I2C.i2c-1.96"]]},
           :permanent, 5000, :worker, [ATECC508A.Transport.I2CServer]}},
         :infinity
       ]}}}}}}}

```
Trying to establish connection to ATEC608A results in similar error:
```
iex(3)> {:ok, i2c} = ATECC508A.Transport.I2C.init(bus_name: "i2c-1")
** (exit) exited in: GenServer.call(ATECC508A.Transport.I2CSupervisor, {:start_child, {{ATECC508A.Transport.I2CServer, :start_link, [["i2c-1", 96, :"Elixir.ATECC508A.Transport.I2C.i2c-1.96"]]}, :permanent, 5000, :worker, [ATECC508A.Transport.I2CServer]}}, :infinity)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
    (elixir 1.11.2) lib/gen_server.ex:1017: GenServer.call/3
    lib/atecc508a/transport/i2c.ex:43: ATECC508A.Transport.I2C.init/1
```

[...]

PR #55, opened 2020-12-04
Work around failing to start nerves_key when optional dep

jjcarstens:
In response to https://github.com/nerves-hub/nerves_hub_link/issues/52


fhunleth:
How about putting the `Application.ensure_all_started` in the NervesKey configurator? Then there's no need for the `if`.

fhunleth (approved):

PR #65, opened 2021-02-24
Add an optional callback for `reboot` in `Client` behaviour

ConnorRigby:
This allows for local development to stub out rebooting


lib/nerves_hub_link/client.ex
```diff
@@ -76,6 +76,14 @@ defmodule NervesHubLink.Client do
   """
   @callback handle_error(any()) :: :ok
 
+  @doc """
+  Optional callback that may be supplied to allow
+  the `reboot` call to be replaced
```
fhunleth, line 81:
```suggestion
  Optional callback to reboot the device when a firmware update completes
  
  The default behavior is to call `Nerves.Runtime.reboot/0` after a successful update. This
  is useful for testing and for doing additional work like notifying users in a UI that a reboot
  will happen soon. It is critical that a reboot does happen.
```

lib/nerves_hub_link/client.ex
```diff
@@ -76,6 +76,14 @@ defmodule NervesHubLink.Client do
...
+  """
+  @callback reboot() :: no_return()
+
+  @optional_callbacks [reboot: 0]
```
jjcarstens, line 88:
```suggestion
  @callback handle_reboot() :: no_return()

  @optional_callbacks [handle_reboot: 0]
```

I would prob say this should be handled a bit like the other callbacks.

Also, do we catch any error and potentially force the reboot anyway?

fhunleth (approved):

jjcarstens (commented):
I'm just a tidbit uneasy leaving the reboot in the hands of the user. If someone takes over this callback and never actually reboots, then device could be "bricked" without manual intervention. But I see what you're trying to do and not really another suggestion of how to accomplish it.

lib/nerves_hub_link/client.ex
ConnorRigby: i don't really think that's necessary personally. I'm thinking that if a user implements their own `reboot` function, and they mess it up, it's either
1) in development, reboot would not be wanted if i were this developer
2) bad firmware that should crash and cause a quarantine anyway

ConnorRigby: differ to @fhunleth what do you think?

fhunleth: There's no error to catch since this has to happen in a separate process. 

Regarding misuse, the `fwup` configs can be bulletproofed to protect against bricking devices via double-upgrading. Some of them do, but it's one of those things that you have to think about and `fwup` is dumb by design on that aspect. I think this is ok.

I don't have opinions on handling the callback implementation. I do like consistency.

ConnorRigby: I have one other update that i'd like to get in, so i'll add it for consistency.

jjcarstens (approved):

jjcarstens (approved):

lib/nerves_hub_link/client.ex
```diff
@@ -127,6 +138,22 @@ defmodule NervesHubLink.Client do
     end
   end
 
+  @doc """
+  Spawns a reboot. This allows the process that
+  calls this function to return before the reboot actually
+  starts.
+  """
+  def spawn_reboot do
```
fhunleth, line 146: This needs specs and `()` since it's a public function and we've gone with the all-functions-have-parens-style.

lib/nerves_hub_link/client.ex
```diff
@@ -136,6 +164,7 @@ defmodule NervesHubLink.Client do
   end
 
   # Catches exceptions and exits
+
```
fhunleth, line 167:
```suggestion
```

PR #66, opened 2021-02-24
Set a default `fwup_devpath` in configurator

ConnorRigby:


jjcarstens (approved):

lib/nerves_hub_link/configurator.ex
```diff
@@ -13,11 +13,26 @@ defmodule NervesHubLink.Configurator do
               remote_iex: false,
               socket: [],
               ssl: []
+
+    @type t() :: %__MODULE__{
+            device_api_host: String.t(),
+            device_api_port: String.t(),
+            device_api_sni: charlist(),
+            fwup_public_keys: [binary()],
+            fwup_devpath: String.t(),
```
fhunleth, line 22:
```suggestion
            fwup_devpath: Path.t(),
```

lib/nerves_hub_link/configurator.ex
```diff
@@ -93,4 +109,9 @@ defmodule NervesHubLink.Configurator do
 
     %{config | fwup_public_keys: config.fwup_public_keys ++ fwup_public_keys}
   end
+
+  @spec maybe_add_fwup_devpath(Config.t()) :: Config.t()
+  defp maybe_add_fwup_devpath(config) do
```
fhunleth, line 114: This doesn't feel like a "maybe" unless I'm missing something.

lib/nerves_hub_link/configurator.ex
ConnorRigby: Path.t() is `iodata` which i don't think would actually work

lib/nerves_hub_link/configurator.ex
ConnorRigby: Whoops, that was supposed to be `put_new` . just a sec

lib/nerves_hub_link/configurator.ex
ConnorRigby: i'm totally wrong. we just updated teh other lib to support this

jjcarstens (approved):

PR #67, opened 2021-02-24
Update info

ConnorRigby:
Don't merge until #66 and #65 are merged


lib/nerves_hub_link/device_channel.ex
```diff
@@ -106,7 +113,14 @@ defmodule NervesHubLink.DeviceChannel do
   def terminate(_reason, _state), do: NervesHubLink.Connection.disconnected()
 
   defp handle_join_reply(%{"firmware_url" => url} = update) when is_binary(url) do
-    UpdateManager.apply_update(update)
+    case NervesHubLinkCommon.Message.UpdateInfo.parse(update) do
+      %NervesHubLinkCommon.Message.UpdateInfo{} = info ->
```
fhunleth, line 117:
```suggestion
      {:ok, info} ->
```

In the other place, `parse` returns an `{:ok, info}`, so I assume this would here.

lib/nerves_hub_link/device_channel.ex
```diff
@@ -71,8 +71,15 @@ defmodule NervesHubLink.DeviceChannel do
   end
 
   def handle_info(%Message{event: "update", payload: update}, state) do
-    _ = UpdateManager.apply_update(update)
-    {:noreply, state}
+    case NervesHubLinkCommon.Message.UpdateInfo.parse(update) do
+      {:ok, %NervesHubLinkCommon.Message.UpdateInfo{} = info} ->
```
fhunleth, line 75: It really feels like we should be able to trust that if `UpdateInfo.parse/1` returns `:ok`, that it does return an `UpdateInfo` struct. Are you worried about it not doing that?

ConnorRigby: no, i just always put the struct name so the reader doesn't have to go look at the docs of that function.

lib/nerves_hub_link/device_channel.ex
ConnorRigby: oops!

jjcarstens (approved):

fhunleth (approved):

PR #68, opened 2021-02-26
Use Slipstream as WS client

jjcarstens:
Changes to use Slipstream as the WS client to handle connections with NervesHub web. The intent will be to deprecate `phoenix_client` in favor of this implementation.

So far this is working well. Its been released as `0.10.0-rc.0` so others can test as well


mix.exs
```diff
@@ -103,7 +103,7 @@ defmodule NervesHubLink.MixProject do
       {:nerves_key, "~> 0.5", optional: true},
       {:nerves_runtime, "~> 0.8"},
       {:nerves_hub_link_common, "~> 0.1.3"},
-      {:phoenix_client, "~> 0.11"},
+      {:slipstream, "~> 0.5"},
```
fhunleth, line 106:
```suggestion
      {:slipstream, "~> 0.6.0"},
```
To use the version of slipstream that doesn't pull in phoenix.

dialyzer-ignore.exs
```diff
@@ -0,0 +1,3 @@
+[
+  {"lib/slipstream.ex"}
```
fhunleth, line 2: How is this needed? This seems like a stranger than normal dialyzer workaround.

mix.exs
```diff
@@ -58,7 +58,8 @@ defmodule NervesHubLink.MixProject do
   defp dialyzer() do
     [
       flags: [:race_conditions, :error_handling, :underspecs, :unmatched_returns],
-      plt_add_apps: [:atecc508a, :nerves_key, :nerves_key_pkcs11],
+      plt_add_apps: [:atecc508a, :nerves_key, :nerves_key_pkcs11, :inets, :slipstream],
```
fhunleth, line 61: `:slipstream` should be pulled in automatically via the deps so it really shouldn't be needed here. I'm also surprised that `:inets` needed to be added. That seems like it should be in `:included_applications` so that it's definitely loaded/initialized at runtime before this.

dialyzer-ignore.exs
jjcarstens: https://github.com/NFIBrokerage/slipstream/pull/29

lib/nerves_hub_link/device_channel.ex
```diff
@@ -1,130 +0,0 @@
...
-  @impl GenServer
-  def handle_info(%Message{event: "reboot"}, state) do
-    Logger.warn("Reboot Request from NervesHubLink")
-    Channel.push_async(state.channel, "rebooting", %{})
-    Client.initiate_reboot()
-    {:noreply, state}
-  end
-
-  def handle_info(%Message{event: "update", payload: update}, state) do
-    case NervesHubLinkCommon.Message.UpdateInfo.parse(update) do
```
ConnorRigby, line 74: this didn't make it after the rebase

lib/nerves_hub_link/device_channel.ex
```diff
@@ -1,130 +0,0 @@
...
-
-  def handle_info(_message, state) do
-    {:noreply, state}
-  end
-
-  @impl GenServer
-  def terminate(_reason, _state), do: NervesHubLink.Connection.disconnected()
-
-  defp handle_join_reply(%{"firmware_url" => url} = update) when is_binary(url) do
-    case NervesHubLinkCommon.Message.UpdateInfo.parse(update) do
```
ConnorRigby, line 119: same here

fhunleth:
To keep everyone in the loop, the `gun` dependency is currently a nonstarter. `gun` has very specific version dependencies on `cowlib`. Unfortunately, `cowboy` also has very specific and different version dependencies on `cowlib`. Therefore you can't have `gun` and `cowboy` in the same project. This would break every Nerves project that using Phoenix right now.

I've been looking around for suggestions on alternative websocket libraries or if anyone is interested in working with Loic and whomever maintains his libraries on hex to resolve this, that would be most appreciated.

the-mikedavis:
:wave: @jjcarstens, @fhunleth

Slipstream is changing its low-level WebSocket client from `:gun` to `Mint.WebSocket` (see https://github.com/NFIBrokerage/slipstream/pull/38) in version 0.8.0. We have it on an rc tag and we're taking it for a spin in production but it should get merged pretty soon. It looks like this project already has Mint in the dependencies so it should hopefully take only a minimal effort to update.

The only thing I see in this PR that has changed in Slipstream over the last couple versions is that `:gun_open_options` is now `:mint_opts` (which get passed to `Mint.HTTP.connect/4`) and you should be able to configure the TLS certs there.

I'll let y'all know when v0.8.0 gets published!

fhunleth:
@the-mikedavis This is great news! Thanks for the update!

the-mikedavis:
Happy friday: v0.8.0 is out!

We finished upgrading slipstream in all of our services and it seems to be going well so far. Hopefully the Mint.WebSocket change works out well for y'all as well :slightly_smiling_face:

jjcarstens:
Yay!!! I'm so excited for this!

I hope to find time soon to try this out and get slipstream working here!

❤️ 💜 💕 🎉

fhunleth (approved):

PR #69, opened 2021-03-01
Track currently downloading firmware UUID

ConnorRigby:


jjcarstens:
What if we move `UpdateInfo` storage into the `UpdateManager` state? Then we can ask the update manager for the UUID we're updating to?

PR #70, opened 2021-03-01
Implement NervesHubLink.Client.handle_fwup_message/2

jjcarstens:
This allows us to pass firmware meta data about the fwup message to the client as well as the web.

Alternate to #69


lib/nerves_hub_link/client/default.ex
```diff
@@ -12,24 +12,29 @@ defmodule NervesHubLink.Client.Default do
...
   end
 
-  def handle_fwup_message(fwup_message) do
-    Logger.warn("Unknown FWUP message: #{inspect(fwup_message)}")
+  def handle_fwup_message(fwup_message, meta) do
+    Logger.warn("Unknown FWUP message [#{meta[:uuid]}]: #{inspect(fwup_message)}")
+  end
+
+  @impl NervesHubLink.Client
+  def handle_fwup_message(_fwup_message) do
```
fhunleth, line 36: This can be removed, right?

jjcarstens: Yes. Forgot that was in there 🤦

ConnorRigby:
I like this much better than #69. still pondering on it a little bit.

jjcarstens:
Changing the client behavior is just a bonus. But we could probably do without it, or punt until there is a need. That would greatly simplify this PR

ConnorRigby (approved):

ConnorRigby:
> Changing the client behavior is just a bonus. But we could probably do without it, or punt until there is a need. That would greatly simplify this PR

I like the behaviour change, especially since it's backwards compatible.

ConnorRigby:
This turned out to not be necessary. See #71

PR #71, opened 2021-03-01
Update DeviceChannel join to contain the current download

ConnorRigby:


fhunleth (approved):

lib/nerves_hub_link/application.ex
```diff
@@ -18,16 +18,16 @@ defmodule NervesHubLink.Application do
     fwup_config = %FwupConfig{
       fwup_public_keys: config.fwup_public_keys,
       fwup_devpath: config.fwup_devpath,
-      handle_fwup_message: &Client.handle_fwup_message/1,
+      handle_fwup_message: &Client.handle_fwup_message/2,
```
jjcarstens, line 21: Dont we still need to change the client here? This arity does not currently exist

PR #74, opened 2021-03-05
v0.10.0 release

jjcarstens:


ConnorRigby (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
+
+* Enhancements
+  * Report `device_api_version` in web connections
+  * Report `currently_downloading_uuid` when joining channel to aid
+    in resumable downloads
+  * Optional `NervesHubLink.Client.reboot/0` callback behavior
+
+* Fixes
+  * Check `firmware_url` is valid before attempting update
+  * set default `fwup_devpath` in the config
```
fhunleth, line 15: I think I may have missed reviewing the `fwup_devpath` or didn't appreciate it. Setting a default is probably not a feature, since if it ever got used, it would mask a really serious issue.

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
 
+## 0.10.0
+
+* `nerves_hub_link_common 0.2.0`
+
+* Enhancements
+  * Report `device_api_version` in web connections
+  * Report `currently_downloading_uuid` when joining channel to aid
+    in resumable downloads
+  * Optional `NervesHubLink.Client.reboot/0` callback behavior
```
fhunleth, line 11: How about
Perhaps this?

```suggestion
* Enhancements
  * Report API version to NervesHub. While there's currently only one version of the Device API, this is anticipation that we may want to change it in the future.
  * On reconnect, notify NervesHub of firmware currently being downloaded so that NervesHub can differentiate failed firmware updates from network hiccups.
  * Allow devices to implement their own reboot logic by implementing the `NervesHubLink.Client.reboot/0` callback
```

fhunleth (approved):
Could you word wrap? Otherwise, it looks good to me.

PR #83, opened 2021-04-09
Allow supplying custom CA certificates to `NervesHubLink`

ConnorRigby:

PR #85, opened 2021-04-16
Release 0.11.0

ConnorRigby:


fhunleth:
I haven't been following closely, but if someone blindly updates to v0.11.0 who's using their own nerves_hub instance, is it possible for them to make a firmware that no longer connects?

ConnorRigby:
> I haven't been following closely, but if someone blindly updates to v0.11.0 who's using their own nerves_hub instance, is it possible for them to make a firmware that no longer connects?

If by blindly, you mean they ignore all the warnings that get printed then yes i think that could happen.

fhunleth:
Is there any reason that the warnings can't be fatal and break builds?

jjcarstens:
If they are connecting to their own instance, then they are either using the `config :nerves_hub_link, ca_certs: ca` option and this will totally fail:
https://github.com/nerves-hub/nerves_hub_link/blob/main/lib/nerves_hub_link/certificate.ex#L29-L47

Or they are using a configurator to supply their own cacerts, in which case it would still connect

fhunleth:
Cool. Thanks!

ConnorRigby:
> Is there any reason that the warnings can't be fatal and break builds?

Jon and I tinkered with this. If there is a situation that can be detected, we raise [here](https://github.com/nerves-hub/nerves_hub_link/pull/83/files#diff-e6adb5a3bf0a85b242d4b77d3100a13fafe420f8f9efdd75fa26ecb2f5cf724cR29)

But we can't detect all situations due to certs being loaded at runtime not compile time

fhunleth (approved):

jjcarstens (approved):

PR #90, opened 2021-08-12
add NervesHubLink.reconnect/0

jjcarstens:
For the cases you might want to bump the socket and restart connections to NH (such as network interface changing, etc)

### DIFF
--- a/lib/nerves_hub_link.ex
+++ b/lib/nerves_hub_link.ex
@@ -1,4 +1,7 @@
 defmodule NervesHubLink do
+  alias NervesHubLink.Supervisor, as: NHLSupervisor
+  alias NervesHubLink.{ConsoleChannel, DeviceChannel, Socket}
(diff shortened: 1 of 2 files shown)


lib/nerves_hub_link.ex
```diff
@@ -1,4 +1,7 @@
 defmodule NervesHubLink do
+  alias __MODULE__.Supervisor, as: NHLSupervisor
```
ConnorRigby, line 2: any reason to use `__MODULE__` here instead of just using the module's name? i don't have any serious opinions on it, but it's not something i see often.

ConnorRigby (approved):

lib/nerves_hub_link.ex
jjcarstens: Nope. Just happen I guess

PR #99, opened 2021-12-07
Jittered connect backoff

jfcloutier:
Computes a jittered list of 11 exponentially growing  backoff values﻿ ending at around 60 secs

### DIFF
--- a/lib/nerves_hub_link/socket.ex
+++ b/lib/nerves_hub_link/socket.ex
@@ -51,7 +51,8 @@ defmodule NervesHubLink.Socket do
     opts = [
       mint_opts: [protocols: [:http1], transport_opts: config.ssl],
       uri: config.socket[:url],
-      rejoin_after_msec: [@rejoin_after]
+      rejoin_after_msec: [@rejoin_after],
+      reconnect_after_msec: backoff()
     ]
 
     socket =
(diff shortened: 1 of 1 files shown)


grace-in-wonderland (approved):

jjcarstens (approved):

lib/nerves_hub_link/socket.ex
```diff
@@ -51,7 +51,8 @@ defmodule NervesHubLink.Socket do
     opts = [
       mint_opts: [protocols: [:http1], transport_opts: config.ssl],
       uri: config.socket[:url],
-      rejoin_after_msec: [@rejoin_after]
+      rejoin_after_msec: [@rejoin_after],
```
jjcarstens, line 54: Do you think we should apply the same jitter for rejoining channels as well?

jfcloutier: Why not? However the default values are different between the two

```
reconnect_after_msec: [
      ...
      default: [10, 50, 100, 150, 200, 250, 500, 1_000, 2_000, 5_000]
    ],
    rejoin_after_msec: [
      ...
      default: [100, 500, 1_000, 2_000, 5_000, 10_000]
    ],
```
I wondered how finely tuned these default values are, i.e. should they have different jittered backoffs or the same?

jfcloutier: Are channels vulnerable to thundering herds, i.e. does jittering matter?

jjcarstens: Unsure on both counts actually. I would imagine reconnect is more intensive than rejoining a channel which is essentially a simple socket message

PR #102, opened 2021-12-07
Change the exponential backoff to take a min and max value

fhunleth:
While it's hardcoded now, it will be possible to alter the backoff to
start and end it at specified values. This is hardcoded to start at a 1
second delay and progressively double until a 60 second delay.

Jitter is specified as a percentage to add. For 0.25 jitter, this means
that would normally be a 1 second delay might be up to 1.25 seconds.
Similarly, a 10 second delay might be as long as 12.5 seconds.

### DIFF
--- a/lib/nerves_hub_link/backoff.ex
+++ b/lib/nerves_hub_link/backoff.ex
@@ -1,0 +1,44 @@
+defmodule NervesHubLink.Backoff do
+  @moduledoc """
+  Compute retry backoff intervals used by Slipstream
+  """
+
+  @doc """
+  Produce a list of integer backoff delays with jitter
+
+  The first two parameters are minimum and maximum value. These are expected to
+  be milliseconds, but this function doesn't care. The returned list will start
+  with the minimum value and then double it until it reaches the maximum value.
+
+  The third parameter is the amount of jitter to add to each delay. The value
+  should be between 0 and 1. Zero adds no jitter. A value like 0.25 will add up
+  to 25% of the delay amount.
+  """
+  @spec delay_list(integer(), integer(), number()) :: [integer()]
+  def delay_list(min, max, jitter) when min > 0 and max >= min and jitter >= 0 do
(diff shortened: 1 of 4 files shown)


jjcarstens (approved):

fhunleth:
@jjcarstens I updated the unit test. Could you look at it again?

jjcarstens (approved):

fhunleth:
@jjcarstens I'm going to run the backoff and jitter settings by a few more people.

jfcloutier (approved):

rraub (approved):
I think these backoff and jitter values are a really good starting point, we don't want to be too quick with retries since actually making the connection eventually is more important than doing it really quickly.

fhunleth:
The following came out after more offline discussion:

1. Need more jitter. 0.50 seems good so the worst case reconnect interval will be 90 seconds. The mean worst case retry interval will be 75 seconds. The desire is to spread out the reconnect attempts more and there didn't seem to be any downside to the added time. I.e., would you notice a difference of 30 seconds to get a firmware update or connect to a shell console if the hub has failed to connect so many times before?
2. Verify that the jitter was random and that that the random seed wasn't based on a timestamp.

fhunleth:
Turns out the seed is based on a timestamp by default: https://github.com/erlang/otp/blob/maint/lib/stdlib/src/rand.erl#L269-L271

fhunleth:
@jjcarstens I think it's ready now. I moved the code around so that the backoff code is in one module. I did that since `socket.ex` just had way too much stuff in it. Aside from the use of a strong random number to seed the jitter values, code mostly just moved around. 

I tested this on a real device. On boot, there are two timeouts right out of the gate since requests are made before the network is available. Prior to the change, there were a flurry of connect attempts since the starting reconnect delays were all subsecond.

jjcarstens (approved):

PR #103, opened 2021-12-07
Fix warnings

fhunleth:
- Fix deprecation warnings on Mix.Config
- Use default configurator to make dev use work

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,4 +1,4 @@
-use Mix.Config
+import Config
 
 # This config.exs file will configure `nerves_hub_link` to point to a local instance
 # of `nerves_hub_web`. See CONTRIBUTING.md for details.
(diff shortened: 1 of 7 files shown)


jjcarstens (commented):
I think if we require this, we also need to bump the minimum supported elixir version which is currently `1.8`

fhunleth:
Oh my, we still support Elixir 1.8 here? Let me fix that. Elixir 1.9 is getting annoying to support now in other libraries, and I'm not 100% sure it's actually possible to set up a non-toy Nerves device with this library and Elixir 1.8.

fhunleth:
```
==> nerves_hub_ca_store
warning: the dependency :nerves_hub_ca_store requires Elixir "~> 1.11" but you are running on v1.9.4
==> extty
warning: the dependency :extty requires Elixir "~> 1.10" but you are running on v1.9.4
```

I assume that first one could support earlier Elixir versions and there's a good chance on the second one. 

Since we currently only test back to Elixir 1.10 on CI and I'm highly considering deprecating Elixir 1.9 on `nerves`, I'm thinking that we change the requirement to Elixir 1.10. If someone complains, perhaps we can reevaluate.

jjcarstens (approved):

jjcarstens (approved):

PR #118, opened 2023-03-20
Send connection types to NervesHub

oestrich:
Adds a callback to let the device have a hook for knowing that its connected successfully to nerves hub. This allows sending connection types up the socket after fully connected. Connection types are "ethernet", "wifi", "cellular".

There is a matching PR for Nerves Hub to save this information.


fhunleth (commented):
Are there any concerns about this causing issues with NervesHub servers that don't support the connection type?

oestrich:
It crashes the socket if we send a message like this to NervesHub servers that don't support it. But since you have to call it in your own code, is it fine to leave as a note in the docs for `send_connection_types/1`? I think we could make NervesHubLink smarter with NervesHub sending capabilities on connect, but it would be a bit of work to make work how we want.

PR #121, opened 2023-05-17
Stop IEx process with :normal instead of 10_000

bjyoungblood:
Unless there's a reason to stop the IEx process with a non-normal status, this seems like it was probably an oversight.

### DIFF
--- a/lib/nerves_hub_link/socket.ex
+++ b/lib/nerves_hub_link/socket.ex
@@ -299,7 +299,7 @@ defmodule NervesHubLink.Socket do
 
   defp stop_iex(%{assigns: %{iex_pid: iex}} = socket) do
     _ = Process.unlink(iex)
-    :ok = GenServer.stop(iex, 10_000)
+    :ok = GenServer.stop(iex, :normal)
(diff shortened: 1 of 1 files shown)


jjcarstens (approved):
Well...I can't remember if API changed here or if you've always be able to specify a stop reason. The intention was to give 10 seconds for the process to close 🤦

lib/nerves_hub_link/socket.ex
```diff
@@ -299,7 +299,7 @@ defmodule NervesHubLink.Socket do
 
   defp stop_iex(%{assigns: %{iex_pid: iex}} = socket) do
     _ = Process.unlink(iex)
-    :ok = GenServer.stop(iex, 10_000)
+    :ok = GenServer.stop(iex, :normal)
```
fhunleth, line 302: Great find!

```suggestion
    GenServer.stop(iex, :normal, 10_000)
```
I suspect that the intention was to override the timeout on stop so it wouldn't be `:infinity`. I also don't think the `:ok` is needed since I don't think `GenServer.stop/3` can return anything else.

PR #127, opened 2023-06-28
add ability to instance socket connection in order to test multiple connections

ConnorRigby:


lib/nerves_hub_link.ex
```diff
@@ -6,47 +6,47 @@ defmodule NervesHubLink do
...
   @spec send_update_progress(non_neg_integer()) :: :ok
-  defdelegate send_update_progress(progress), to: Socket
+  def send_update_progress(progress), do: Socket.send_update_progress(Socket, progress)
 
   @doc """
   Send an update status to web
   """
   @spec send_update_status(String.t() | atom()) :: :ok
-  defdelegate send_update_status(status), to: Socket
+  def send_update_status(status), do: Socket.send_update_status(Socket, status)
```
jjcarstens, line 51:
```suggestion
  defdelegate send_update_status(socket \\ Socket, status), to: Socket
```
I'm pretty sure you can delegate the default and it would work with dialyzer as well

lib/nerves_hub_link.ex
```diff
@@ -6,47 +6,47 @@ defmodule NervesHubLink do
...
-  defdelegate status(), to: NervesHubLink.UpdateManager
+  def status(), do: NervesHubLink.UpdateManager.status(Socket)
 
   @doc """
   Restart the socket and device channel
   """
-  @spec reconnect() :: :ok
-  defdelegate reconnect(), to: Socket
+  @spec reconnect_socket() :: :ok
+  def reconnect_socket(), do: Socket.reconnect_socket(Socket)
```
jjcarstens, line 39: Is there a specific reason for this change? This could be considered breaking since `reconnect()` has been around a while

lib/nerves_hub_link/configurators/load_test.ex
jjcarstens, line 1: I don't think we need a load test file. This is basic enough that those testing can implement what they need and we can skip adding a file thats not used in production

lib/nerves_hub_link.ex
ConnorRigby: yeah  `use Slipstream` imports a `reconnect/2` function. We could make `NervesHubLink.reconnect` stay the same while the socket module remains the new function.

PR #128, opened 2023-08-14
Add a client callback for changing the backoff timeouts on reconnects

oestrich:
Allow for changing the timeouts between the connection dropping and trying to reconnect.

### DIFF
--- a/lib/nerves_hub_link/client.ex
+++ b/lib/nerves_hub_link/client.ex
@@ -77,6 +77,13 @@ defmodule NervesHubLink.Client do
   @callback handle_error(any()) :: :ok
 
   @doc """
+  Called when the socket disconnected, before starting to reconnect
+
+  The return value is is used to reset the next socket's retry timeout
+  """
+  @callback reconnect_backoff() :: [integer()]
+
+  @doc """
   Callback to identify the device from NervesHub.
   """
   @callback identify() :: :ok
@@ -172,6 +179,19 @@ defmodule NervesHubLink.Client do
   @spec handle_error(any()) :: :ok
   def handle_error(data) do
     _ = apply_wrap(mod(), :handle_error, [data])
+  end
+
+  @doc """
+  This function is called internally by NervesHubLink to notify clients of disconnects.
+  """
+  def reconnect_backoff() do
+    backoff = mod().reconnect_backoff()
+
+    if is_nil(backoff) do
+      NervesHubLink.Backoff.delay_list(1000, 60000, 0.50)
+    else
+      backoff
+    end
   end
 
   # Catches exceptions and exits
--- a/lib/nerves_hub_link/client.ex
+++ b/lib/nerves_hub_link/client.ex
@@ -81,7 +81,7 @@ defmodule NervesHubLink.Client do
 
   The return value is is used to reset the next socket's retry timeout
   """
-  @callback reconnect_backoff() :: [integer()]
+  @callback reconnect_backoff() :: [integer()] | nil
 
   @doc """
   Callback to identify the device from NervesHub.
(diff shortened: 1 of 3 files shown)


lib/nerves_hub_link/client.ex
```diff
@@ -174,6 +178,13 @@ defmodule NervesHubLink.Client do
     _ = apply_wrap(mod(), :handle_error, [data])
   end
 
+  @doc """
+  This function is called internally by NervesHubLink to notify clients of disconnects.
+  """
+  def reconnect_backoff() do
+    apply_wrap(mod(), :reconnect_backoff, [])
```
jjcarstens, line 185: `apply_wrap/3` has the potential to blow up and return error tuples. This would kill any reconnect trying to happen. Should we instead apply a default here as well to give it a fighting chance? Or let that error propagate to a potentially incorrect client?

oestrich: I think applying the backoff if the client returns nil is good 👍

lib/nerves_hub_link/client.ex
```diff
@@ -76,6 +76,13 @@ defmodule NervesHubLink.Client do
   """
   @callback handle_error(any()) :: :ok
 
+  @doc """
+  Called when the socket disconnected, before starting to reconnect
+
+  The return value is is used to reset the next socket's retry timeout
```
fhunleth, line 82:
```suggestion
  The return value is used to reset the next socket's retry timeout. `nil` uses
  the default.
```

lib/nerves_hub_link/client.ex
```diff
@@ -76,6 +76,13 @@ defmodule NervesHubLink.Client do
...
+  """
+  @callback reconnect_backoff() :: [integer()]
```
fhunleth, line 84:
```suggestion
  @callback reconnect_backoff() :: [integer()] | nil
```

lib/nerves_hub_link/client.ex
```diff
@@ -174,6 +182,19 @@ defmodule NervesHubLink.Client do
...
+  This function is called internally by NervesHubLink to notify clients of disconnects.
+  """
+  def reconnect_backoff() do
+    backoff = mod().reconnect_backoff()
+
+    if is_nil(backoff) do
+      NervesHubLink.Backoff.delay_list(1000, 60000, 0.50)
+    else
+      backoff
+    end
```
fhunleth, line 195: I switched to success logic, and I think that I'm for using `apply_wrap/3`. The reason is that if anything goes wrong with getting the `backoff` values, I think that crashing causes an immediate restart and retry. That seems undesirable. I also think that `apply_wrap/3` really should log an error when it catches something (future PR, though, since that not your problem).

```suggestion
    if is_list(backoff) do
      backoff
    else
      NervesHubLink.Backoff.delay_list(1000, 60000, 0.50)
    end
```

lib/nerves_hub_link/client.ex
```diff
@@ -174,6 +182,19 @@ defmodule NervesHubLink.Client do
     _ = apply_wrap(mod(), :handle_error, [data])
   end
 
+  @doc """
+  This function is called internally by NervesHubLink to notify clients of disconnects.
+  """
+  def reconnect_backoff() do
```
fhunleth, line 188:
```suggestion
  """
  @spec reconnect_backoff() :: [integer()]
  def reconnect_backoff() do
```

lib/nerves_hub_link/client.ex
```diff
@@ -174,6 +182,20 @@ defmodule NervesHubLink.Client do
     _ = apply_wrap(mod(), :handle_error, [data])
   end
 
+  @doc """
+  This function is called internally by NervesHubLink to notify clients of disconnects.
```
fhunleth, line 186: Could you add a short paragraph about why someone would want to use this function?

Also, it seems like it should be an optional callback since I would assume that without other guidance, most people would go with the default. Right?

fhunleth (approved):

PR #129, opened 2023-08-22
Bump version to 2.0.0

oestrich:
The new release of NervesHubLink starts adding new features from NervesHub 2.0, such as identifying the device you have open in NervesHub and hooks to help prevent a thundering herd of device reconnects. It also cleans up the code base a bit by removing unused packages. Make sure to run `mix deps.unlock --unused` after updating to keep your lock file up to date.

* Removed
  * NervesHubCLI and NervesHubCAStore as dependencies
  * NervesHubLinkCommon as dependencies, it was merged into this repo

* Added
  * Identify callback from NervesHub, a new message from the server that
  will let you blink an LED or anything else that will help you identify
  the device you're looking at is the one you have open in NervesHub.
  * Add a client callback for changing the backoff timeouts on reconnects (#128)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,29 @@
 # Changelog
 
+## [2.0.0] - 2023-08-22
+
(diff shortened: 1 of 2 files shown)


fhunleth:
Is there a general warm fuzzy that we can give to people? Like this is a safe update if you're using NervesHub 2.0. I'm just thinking of adding one sentence to encourage people to upgrade if they can.

fhunleth (commented):
Approve!

PR #131, opened 2023-10-18
Use the console channel to save files to the device

oestrich:
Device side of console file uploading from the NervesHub remote console

Related: https://github.com/nerves-hub/nerves_hub_web/pull/1059

### DIFF
--- a/lib/nerves_hub_link/socket.ex
+++ b/lib/nerves_hub_link/socket.ex
@@ -2,6 +2,7 @@ defmodule NervesHubLink.Socket do
   @moduledoc false
 
   use Slipstream
+
   require Logger
 
   alias NervesHubLink.Client
@@ -102,6 +103,7 @@ defmodule NervesHubLink.Socket do
       |> assign(iex_pid: nil)
       |> assign(iex_timer: nil)
       |> assign(uploader_pid: nil)
+      |> assign(data_path: config.data_path)
       |> connect!(opts)
 
     Process.flag(:trap_exit, true)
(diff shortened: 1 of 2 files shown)


lib/nerves_hub_link/socket.ex
```diff
@@ -296,6 +298,27 @@ defmodule NervesHubLink.Socket do
...
+    :ok = File.touch!(path)
+    {:ok, socket}
+  end
+
+  def handle_message(@console_topic, "file-data", params, socket) do
+    path = Path.join(socket.assigns.upload_path, params["filename"])
+    file = File.open!(path, [:append])
+    chunk = Base.decode64!(params["data"])
+    :ok = IO.binwrite(file, chunk)
+    :ok = File.close(file)
```
jjcarstens, line 314:
```suggestion
    file = File.open!(path, [:append], fn fd ->
        chunk = Base.decode64!(params["data"])
        IO.binwrite(fd, chunk)
    end)
```

It's prob worth using the function to `File.open!/3` here so that if there is an error it will definitely get closed

lib/nerves_hub_link/socket.ex
```diff
@@ -102,6 +103,7 @@ defmodule NervesHubLink.Socket do
       |> assign(iex_pid: nil)
       |> assign(iex_timer: nil)
       |> assign(uploader_pid: nil)
+      |> assign(upload_path: config.upload_path)
```
jjcarstens, line 106: This might be slightly confusing since there is also an `uploader_pid`, but that is technically completely different. I don't know if I have a better name though.  `data_dir`? `storage_path`? `:downloads`?

oestrich: Good catch. I hate that we're uploading and downloading on both sides lol. `data_dir` sounds like the best choice

jjcarstens (approved):

PR #135, opened 2023-11-03
ArchiveManager to handle NervesHub sending extra firmware

oestrich:
Archives are an extra firmware attached to a deployment. On connect NervesHub will potentially send a message about an archive and the device can download the firmware to a configured / known location and tell the configured client that it's there. ArchiveManager does nothing else as the extra firmware will be highly application specific.

Related: https://github.com/nerves-hub/nerves_hub_web/pull/1091

### DIFF
--- a/lib/nerves_hub_link/archive_manager.ex
+++ b/lib/nerves_hub_link/archive_manager.ex
@@ -1,0 +1,178 @@
+defmodule NervesHubLink.ArchiveManager do
+  @moduledoc """
+  GenServer for handling downloading archives from NervesHub
+
+  Your NervesHubLink client will tell the manager when to download
+  an archive and the Manager will tell your client when it's done
+  downloading so you can act on it.
+
+  You are expected to remove the file when you're done with it and track
+  that it has been applied to prevent downloading again.
+  """
+
+  use GenServer
+
+  alias NervesHubLink.Client
+  alias NervesHubLink.Downloader
+  alias NervesHubLink.Message.ArchiveInfo
+
+  require Logger
+
+  @type status :: :idle | :downloading | :done | :update_rescheduled
+
+  @type t :: %__MODULE__{
+          archive_info: nil | ArchiveInfo.t(),
+          data_path: Path.t(),
+          download: nil | GenServer.server(),
+          file_path: Path.t(),
+          status: status(),
+          update_reschedule_timer: nil | :timer.tref()
+        }
+
+  defstruct archive_info: nil,
+            data_path: nil,
+            download: nil,
+            file_path: nil,
+            status: :idle,
+            update_reschedule_timer: nil
+
+  @doc """
+  Must be called when an archive payload is dispatched from
+  NervesHub. the map must contain a `"url"` key.
+  """
+  @spec apply_archive(GenServer.server(), ArchiveInfo.t()) :: status()
+  def apply_archive(manager \\ __MODULE__, %ArchiveInfo{} = archive_info) do
+    GenServer.call(manager, {:apply_archive, archive_info})
+  end
+
+  @doc """
+  Returns the current status of the archive manager
+  """
+  @spec status(GenServer.server()) :: status()
+  def status(manager \\ __MODULE__) do
+    GenServer.call(manager, :status)
+  end
+
+  @doc """
+  Returns the UUID of the currently downloading archive, or nil.
+  """
+  @spec currently_downloading_uuid(GenServer.server()) :: uuid :: String.t() | nil
+  def currently_downloading_uuid(manager \\ __MODULE__) do
+    GenServer.call(manager, :currently_downloading_uuid)
+  end
+
+  @doc false
+  @spec child_spec(map()) :: Supervisor.child_spec()
+  def child_spec(args) do
+    %{
+      start: {__MODULE__, :start_link, [args, [name: __MODULE__]]},
+      id: __MODULE__
+    }
+  end
+
+  @doc false
+  @spec start_link(GenServer.options()) :: GenServer.on_start()
+  def start_link(args, opts \\ []) do
+    GenServer.start_link(__MODULE__, args, opts)
+  end
+
+  @impl GenServer
+  def init(args) do
+    {:ok, %__MODULE__{data_path: args.data_path}}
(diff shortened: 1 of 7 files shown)


lib/nerves_hub_link/archive_manager.ex
```diff
@@ -0,0 +1,148 @@
...
+    _ = Client.archive_available(state.archive_info)
+    {:noreply, %State{state | archive_info: nil, archive_path: nil, download: nil, status: :idle}}
+  end
+
+  def handle_info({:download, {:error, reason}}, state) do
+    Logger.error("[NervesHubLink] Nonfatal HTTP download error: #{inspect(reason)}")
+    {:noreply, state}
+  end
+
+  # Data from the downloader is sent to fwup
```
fhunleth, line 139:
```suggestion
  # Data from the downloader
```

lib/nerves_hub_link/archive_manager.ex
```diff
@@ -0,0 +1,148 @@
+defmodule NervesHubLink.ArchiveManager do
+  @moduledoc """
+  GenServer responsible for brokering messages between:
+    * an external controlling process
+    * HTTP
+
+  Should be started in a supervision tree
```
fhunleth, line 7: The moduledoc sounds like it was copy/pasted from somewhere else. I was thinking this would talk about getting notifications for archives and downloading them.

One thing I'm missing is how a device determines whether it wants archive. Once it has processed the archive, does the archive stick around and if there's a new one, does the old one get deleted?

lib/nerves_hub_link/archive_manager.ex
```diff
@@ -0,0 +1,148 @@
...
+  """
+  use GenServer
+
+  alias NervesHubLink.Client
+  alias NervesHubLink.Downloader
+  alias NervesHubLink.Message.ArchiveInfo
+
+  require Logger
+
+  defmodule State do
```
fhunleth, line 17: We've switched to not use `State` modules any more for GenServer state. It's mostly stylistic, but there is a cost on the slower devices to every `.beam` file that gets loaded.

lib/nerves_hub_link/configurator.ex
```diff
@@ -3,11 +3,12 @@ defmodule NervesHubLink.Configurator do
   alias __MODULE__.{Config, Default}
   require Logger
 
-  @device_api_version "1.0.0"
+  @device_api_version "2.0.0"
   @console_version "2.0.0"
 
   defmodule Config do
-    defstruct connect: true,
+    defstruct archive_path: "/data/archives/",
```
jjcarstens, line 10: I'm torn on this. I'd rather keep downloads in a consolidated location like`#{data_path}/archives` so all NHL download data is scoped, but I could see clients potentially wanting it somewhere else. 🤔 


I think I'd vote for having it in the consolidated location and not configurable. We wouldn't make any assertion on the file existing or not and instead present the info to `NervesHubLink.Client` to let them decide if it should be downloaded now, later, or acted on (like with firmware updates). What do you think?

lib/nerves_hub_link/archive_manager.ex
```diff
@@ -0,0 +1,148 @@
...
+  end
+
+  @impl GenServer
+  def init(args) do
+    {:ok, %State{archive_path: args.archive_path}}
+  end
+
+  @impl GenServer
+  def handle_call({:apply_archive, %ArchiveInfo{} = info}, _from, %State{} = state) do
+    file_path = Path.join(state.archive_path, info.uuid <> ".fw")
```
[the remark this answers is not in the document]
oestrich: Currently archives _have_ to be a `.fw` file so I think this is fine right? We extract information out of the file on NervesHub web side via fwup so it needs to be in this signed format.

lib/nerves_hub_link/archive_manager.ex
oestrich: Huh, I copied from the update manager so I went with what that had. Should I remove the state module from the other manager as a separate PR then?

lib/nerves_hub_link/archive_manager.ex
oestrich: The archive sticks around currently, the UUID is what blocks the archive manager from downloading the same file over and over. If there's a new one we'll keep the old file and download the new one. I'm not sure we can stream these like the existing firmware manager. Someone mentioned we should have a callback before downloading and let the client determine when we download files and your main application can clean up the old files afterwards then maybe?

lib/nerves_hub_link/archive_manager.ex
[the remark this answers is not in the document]
jjcarstens: Why does it have to be? `.fw` is essentially a made up extension to identify fwup created firmware bundles, but you should be able to do it without it

```
> nerves_hub_web √ % fwup -c -f tmp/archive.conf -o archive_another
> nerves_hub_web √ % fwup -S -s tmp/archives_key.priv -i archive_another -o tmp/archive_another
> nerves_hub_web √ % fwup -m -i tmp/archive_another
meta-product="poser"
meta-description="Test archive"
meta-version="0.0.1"
meta-platform="poser-platform"
meta-architecture="x86_64"
meta-creation-date="2023-11-17T21:50:38Z"
meta-uuid="76a83267-8e83-5778-3038-afde0e603558"
```

That said, I feel like we could use a `.fwa` or some other extension to help separate archives from firmware bundles, but that could come later

lib/nerves_hub_link/archive_manager.ex
```diff
@@ -0,0 +1,178 @@
...
+
+    pid = self()
+
+    case Client.archive_available(info) do
+      :download ->
+        {:ok, download} = Downloader.start_download(info.url, &send(pid, {:download, &1}))
+
+        _ = File.mkdir_p(directory)
+        # delete the old one in case it was a failed download
+        _ = File.rm_rf(file_path)
```
jjcarstens, line 150: Should we be more clever about this? If things did fail to download, we now don't even have the old archive. Maybe we need to download to a `${filepath}.download` and then copy it over to `filepath` when successfully completed?

jjcarstens (approved):

PR #137, opened 2023-12-03
Remove `NervesHubLink.Connection`

jjcarstens:
This was previously an Agent for minimal connection state that was used due to the limitations of `phoenix_client` and socket tracking. However, since the switch to `slipstream` is has been unneeded and virtually unused.

There was also some behavior of checking "uptime" of the connection which is removed with this change in favor of NHL users handling that check with their own logic. Times for process start, connection, and successful join are now saved in the socket instead of the Agent for future use


lib/nerves_hub_link/connection.ex
```diff
@@ -1,124 +0,0 @@
...
-  [nerves-hub.org](https://www.nerves-hub.org) channel changes.
-  Likewise, it is set after a HTTP request fails or succeeds. This makes it
-  useful when you want to consider the connection to
-  [nerves-hub.org](https://www.nerves-hub.org) as part of the overall health
-  of the device and perform explicit actions based on the result, such as using
-  the Erlang [:heart](http://erlang.org/doc/man/heart.html) module to force a
-  reboot if the callback check fails.
-
-  ```
-  # Set a callback for heart to check every 5 seconds. If the function returns anything other than
```
fhunleth, line 20: Oh wow. I don't remember this advice for some reason, but it seems best that it is being deleted. A better check would be if there's been no connection a all for the past 15 minutes or hour or whatever, and then to tie it in with firmware validation.

fhunleth (approved):

PR #161, opened 2024-03-22
Add alarms for common error conditiopns

ConnorRigby:
This adds alarms to 
* UpdateManager
* HTTPFwupStream
* Socket

To allow external subscribers such as [Alarmist](https://github.com/smartrent/alarmist) to consume alarms


README.md
```diff
@@ -447,6 +447,20 @@ config :nerves_hub_link, remote_iex_timeout: 900 # 15 minutes
...
+
+* `NervesHubLink.DownloadFail` 
+  * set: HTTP Download of a .fw file has failed.
+  * clear: HTTP Download has completed.
+* `NerbesHubLink.UpdateFail`
+  * set: fwup error occurred while applying an update.
+  * clear: fwupdate has completed.
+* `NervesHubLink.SocketDisconnect`
+  * set: Socket connection to NervesHub has been closed or disconnected.
+  * clear: Socket connection has been restored.
```
fhunleth, line 462:
```suggestion
* `NervesHubLink.Disconnected`
  * set: An issue is preventing a connection to NervesHub or one just hasn't been made yet
  * clear: Currently connected to NervesHub
* `NervesHubLink.UpdateInProgress` 
  * set: A new firmware update is being downloaded or applied
  * clear: No updates are happening
```

I changed the alarms around a bit. DownloadFail/UpdateFail are more events that alarms. I.e., they happened; they're not ongoing conditions. I changed SocketDisconnect to Disconnected to be more generic. Not that it matters now, but I think the consumer of that alarm only cares whether the NHL<->NH connection works or not.

I think that "UpdateInProgress" is what you're looking for since that would be useful in showing a message to a user that they may want to wait for the new firmware to be applied or perhaps that a reboot was imminent.

fhunleth:
This looks pretty good to me. How about we keep it open a little longer just in case we have regrets? Maybe a week or two.

PR #168, opened 2024-04-18
support nerveshub sending current fwup public keys

joshk:
when the device connects it can now request all current fwup public keys are sent down the wire.

here is the general logic:

1. if there are no firmware signing keys configured on the device then it should request them when connecting
2. if 1 or more firmware signing keys are on the device, then it will not request any updated keys
3. keys can be requested during connection via an opt-in setting on the device

### DIFF
--- a/lib/nerves_hub_link/configurator.ex
+++ b/lib/nerves_hub_link/configurator.ex
@@ -13,6 +13,7 @@ defmodule NervesHubLink.Configurator do
               device_api_port: 443,
               device_api_sni: nil,
               fwup_public_keys: [],
+              request_fwup_public_keys: false,
               archive_public_keys: [],
               fwup_devpath: "/dev/mmcblk0",
               fwup_env: [],
@@ -29,6 +30,7 @@ defmodule NervesHubLink.Configurator do
             device_api_port: String.t(),
             device_api_sni: charlist(),
             fwup_public_keys: [binary()],
+            request_fwup_public_keys: boolean(),
             archive_public_keys: [binary()],
             fwup_devpath: Path.t(),
             fwup_env: [{String.t(), String.t()}],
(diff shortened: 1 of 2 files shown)


fhunleth:
@oestrich and @jjcarstens I had a private exchange with Josh on the security implications and wanted to share my thoughts here. I feel this update is fine since I can do everything Josh describes already, but it's harder. Therefore, this doesn't lessen security in my opinion. 

I don't have opinions on how signing keys get transferred from NH to the device. This way seems good, but I figured you both would care more on that aspect.

oestrich:
I don’t think I like being able to change the fwup keys on connect to Nerveshub with something built in. This feels like it’s making it too easy to put on a rogue firmware. But maybe it’s fine if we continue expanding roles 🤷 I’m gonna let this stew for a bit to think it over if that’s ok

jjcarstens:
I think at this level, the battle is already lost if a rogue firmware successfully connects. At that level, we're already trusting the connection and these are public keys we're providing.

Though maybe we make this an opt in on NervesHub to decide if it wants to push keys or not and the link just needs to know how to deal with the message

fhunleth:
I think this is an access control issue. Anyone with IEx prompt access to a device can install whatever firmware they want. To prevent someone with IEx access, we'd have to really lock down that public key and I'm not quite sure how I could even make it hard to circumvent for anyone with Elixir and Linux knowledge. 

If you all want a setting or some sort of set of dialogs that you need to click through or something for the case of replacing a public key, that seems fine to me. Per Josh's point, if a device has no public key, then sending one down seems fine.

oestrich:
I think my concerns are assuaged if we make it opt in for downloading if there are keys present. If someone gets access to nerveshub they can adjust the onconnect code (or console it) and get around all of this anyways, so I think my previous concern is kinda moot. 

I like your bullet points and think that's the way to go 👍

joshk:
@oestrich thanks for your feedback!

@oestrich @fhunleth @jjcarstens I've updated the PR (and web PR) to match the logic in my last comment

I'll update the readme in a different PR as I'm working on a larger update which includes a few extra bits

jjcarstens (approved):

PR #169, opened 2024-04-18
`nerves_time` is essentially a dep, so lets include it

joshk:
link talks to hub, hub uses ssl and certs, so time is pretty much required.

lets just include it by default.


joshk:
@jjcarstens can i haz review pls

jjcarstens (approved):

PR #176, opened 2024-04-23
Make shared secret auth the default

joshk:
along with removing the provisioning code
and simplifying the cert key auth configurator
oh, and also move the 'workflow' section to the cli docs, as they are cli specific

In short, the top tier of security and 'connecting your device to NervesHub` is to use a crypto chip (NervesKey). NervesKey is the 'production grade' deployment setup. If you aren't using NervesKey, then using Shared Secret auth is probably the next option to consider as your device isn't likely out in the field yet.

This is based on some discussions I had with @jjcarstens today.


lib/nerves_hub_link/configurator.ex
```diff
@@ -118,11 +121,13 @@ defmodule NervesHubLink.Configurator do
...
 
   defp fetch_default() do
-    if Code.ensure_loaded?(NervesKey) do
+    default = if Code.ensure_loaded?(NervesKey) do
       NervesHubLink.Configurator.NervesKey
     else
-      Default
+      NervesHubLink.Configurator.SharedSecret
     end
+    Application.put_env(:nerves_hub_link, :configurator, default)
```
jjcarstens, line 129: Why are we changing to put this back into the Application config?

lib/nerves_hub_link/configurators/cert_key.ex
```diff
@@ -1,11 +1,8 @@
-defmodule NervesHubLink.Configurator.Default do
+defmodule NervesHubLink.Configurator.CertKey do
```
jjcarstens, line 1:
```suggestion
defmodule NervesHubLink.Configurator.CertFile do
```
Maybe we indicate this is local file based? (or some other name to indicate that we'll look for a file in a default spot)

lib/nerves_hub_link/configurators/cert_key.ex
```diff
@@ -23,28 +20,11 @@ defmodule NervesHubLink.Configurator.Default do
   end
 
   defp maybe_add_cert(socket_opts) do
-    if socket_opts[:cert] || socket_opts[:certfile] do
-      # option already provided
-      socket_opts
-    else
-      cert =
-        Nerves.Runtime.KV.get(@cert)
```
jjcarstens, line 31: Should we keep this for backwards compatibility? I guess I don't know how much this is used. We can remove the mechanism to try and help put this there, but anyone who has already provisioned through the UBoot vars will be broken

lib/nerves_hub_link/configurator.ex
joshk: It's because of https://github.com/nerves-hub/nerves_hub_link/blob/main/lib/nerves_hub_link/socket.ex#L424-L425
And since there are two defaults, we need to know which was is being used.

lib/nerves_hub_link/configurators/cert_key.ex
joshk: I guess we could, but maybe add logger warnings that this feature of provisioning is no longer supported in link?

I do like the concept of what provisioning does and how it works, but its current setup is a little brittle.

It would be nice if any `NERVES_` env var found during firmware burning was then converted to a uboot provisioning line. That would make it more flexible and also allow for us to remove the `provisioning.conf` file, etc.

joshk: This would also allow for the shared secret auth configs to be stored in env vars and outside of the config, which I really like.

lib/nerves_hub_link/configurators/local_cert_key.ex
```diff
@@ -0,0 +1,58 @@
...
+      socket_opts[:cert] || socket_opts[:certfile] ->
+        socket_opts
+
+      cert = Nerves.Runtime.KV.get(@cert_kv_path) ->
+        cert = Certificate.pem_to_der(cert)
+
+        Keyword.put(socket_opts, :cert, cert)
+
+      true ->
+        Keyword.put_new(socket_opts, :certfile, "/data/nerves_hub/cert.pem")
```
joshk, line 36: change this to use the data path in the config as the path default

lib/nerves_hub_link/configurator.ex
```diff
@@ -39,6 +40,8 @@ defmodule NervesHubLink.Configurator do
             nerves_key: any(),
             params: map(),
             remote_iex: boolean(),
+            request_fwup_public_keys: boolean(),
+            shared_secret: [{String.t(), String.t()}],
```
jjcarstens, line 44:
```suggestion
            shared_secret: [product_key: String.t(), product_secret: String.t()],
```
These are currently expected as atoms, correct?

lib/nerves_hub_link/configurators/local_cert_key.ex
```diff
@@ -0,0 +1,62 @@
...
+        ssl[:cert] || ssl[:certfile] ->
+          ssl
+
+        cert = Nerves.Runtime.KV.get(@cert_kv_path) ->
+          cert = Certificate.pem_to_der(cert)
+
+          Keyword.put(ssl, :cert, cert)
+
+        true ->
+          Keyword.put_new(ssl, :certfile, "#{config.data_path}/cert.pem")
```
jjcarstens, line 34:
```suggestion
          Keyword.put_new(ssl, :certfile, Path.join(config.data_path, "cert.pem")
```
Probably worth using the path helper here just for cleanliness so we don't end up with double slashes (`/data/nh//cert.pem`, etc)

jjcarstens (commented):
Found just a couple little things

fhunleth (approved):
Looks good to me at a high level. I'm certainly good with shared secret being the new default.

jjcarstens (approved):

PR #177, opened 2024-04-25
Have link wait some number of seconds before connecting

joshk:
It usually takes 10-20 seconds for the network to become available after the device is started. 
This adds a 15-second delay and can be configured to whatever is preferred

### DIFF
--- a/lib/nerves_hub_link/configurator.ex
+++ b/lib/nerves_hub_link/configurator.ex
@@ -8,6 +8,7 @@ defmodule NervesHubLink.Configurator do
 
   defmodule Config do
     defstruct connect: true,
+              connect_wait: 15_000,
               data_path: "/data/nerves-hub",
               device_api_host: nil,
               device_api_port: 443,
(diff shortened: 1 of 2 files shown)


joshk:
@jjcarstens expressed similar concerns.

The thing to remember is that the retry backoff after a disconnect, e.g. due to a broken network connection, will climb like so 1, 2, 4, 8, 16, 32, 60. This PR is only about delaying the first connection try on boot, and its configurable. 

What I saw when onboarding some new users to NervesHub is the device (a raspberry pi) took about 10-15 seconds for network connectivity to be available and during that time there were a slew log messages and connectivity issues, and due to the retry backoff, we could then ssh into the device before Link would try and connect again.

I think a likely better approach would be to wait for `:inet.getaddr("nerveshub.host.url")` to return `{:ok, _}` before trying to connect, and polling that every second or two.

fhunleth:
You're right about the backoff algorithm. I like your idea of waiting until DNS starts working to start the connect, and I saw #181.

This makes me want to explore having Slipstream back off differently depending on what error it gets back from connecting. For example, `:nxdomain` wouldn't back off like what you've done here and that way the reconnect timeouts work the same regardless of whether it's right after boot or not. I think this needs some more thought, and it involves modifying Slipstream, so nothing I want to do now since you addressed the most important case, imho.

joshk:
Thanks for the feedback @fhunleth.

I'm going to close this PR as its superseded by #181

PR #178, opened 2024-04-25
The `Client` behaviour can now be used with `use`

joshk:
This allows people to use the defaults and override the implementations they want without having to implement all of them. 
The `ClientTest` no longer really makes sense, so I've removed it.
This does remove the function value checking that `apply_wrap` was doing, although I think its fair that this be a responsibility of people implementing the behaviour or overriding the default.


lib/nerves_hub_link/client.ex
```diff
@@ -118,139 +122,116 @@ defmodule NervesHubLink.Client do
...
     end
   end
-
-  # Catches exceptions and exits
-  defp apply_wrap(mod, function, args) do
-    apply(mod, function, args)
-  catch
-    :error, reason -> {:error, reason}
-    :exit, reason -> {:exit, reason}
-    err -> err
```
jjcarstens, line 250: I'd actually say this is _the_ feature of the client. Thw whole purpose here was to guard the user from doing stupid things that completely severe the connection to NervesHub and leave their device in a bricked state unable to get a remote update.

Switching to the `use` model makes sense, but I think we need to find a way to have this protection still. With the proposed changes, every direct call on the client in the socket has the potential to blow up and kill it

lib/nerves_hub_link/client.ex
```diff
@@ -118,139 +122,116 @@ defmodule NervesHubLink.Client do
...
 
-  @doc """
-  This function is called internally by NervesHubLink to notify clients.
-  """
-  @spec update_available(update_data()) :: update_response()
-  def update_available(data) do
-    case apply_wrap(mod(), :update_available, [data]) do
-      :apply ->
-        :apply
+      @impl true
```
jjcarstens, line 134:
```suggestion
      @impl NervesHubLink.Client
```
By convention, I like to be explicit in the impl for more clarity and extra compiler help

joshk: I'll change these back

lib/nerves_hub_link/client.ex
joshk: nerves is smart with rollbacks to previous versions, would that not possibly help?

could we do the smart logic in the areas that the methods are called?

lib/nerves_hub_link/client.ex
joshk: updated

lib/nerves_hub_link/client.ex
jjcarstens: Nerves can be smart with rollbacks, but it's only default in `bbb`. Most systems need to implement that logic/behavior.

And yes, totally can do the smart logic in the areas. Sorry my comment wasn't totally clear but the intention was definitely "there was a reason we did this, so lets make sure we think of that reason where things are being called"

joshk: totally get cha

maybe everywhere one of these methods are called we wrap that call instead? we could call it `trap_errors`?

PR #187, opened 2024-05-12
Archive public key defaults

joshk:
This simplifies the setup of public keys for archive verification purposes.

If there are no public keys setup for archive verification defined...
1. use the `fwup_public_keys`, if they exist
2. ask `NervesHub` to send over the latest public keys

Also support the ability in the config to request the latest public keys from `NervesHub` even if `fwup_public_keys` are defined in the config.

Related: https://github.com/nerves-hub/nerves_hub_web/pull/1291

**Please Note:** 

I made a significant change to `UpdateManager` and `ArchiveManager` where they don't store the public verification keys when they are started, but are instead passed them by the socket when an archive or update is ready to be downloaded. 

This is because if the keys were updated, but then the process died, they would not have the most up to date keys.


oestrich (commented):
This was purposefully split as separate concepts, I don't think we want to merge them. They might be the same but at least for smartrent they're separate and we do not want firmware keys to be signing archives and vice versa

joshk:
@oestrich I've updated the PR based on our chat.

jjcarstens (approved):

Issue #47, opened 2020-09-30
Request new firmware URL if current is expired


jjcarstens:
Moving this from https://github.com/nerves-hub/nerves_hub_web/issues/595

TL;DR - If you get a firmware url and delay applying it before the TTL expires (~10 minutes), then issues happening.

We should adapt the download request here to ask for a new, generated URL if the link is expired, but limit to only X number of retries (3 maybe?)

This will prob require new channel message in `nerves_hub_web` to support getting another download URL because currently it only happens on channel join _or_ when deployment/device changes in `nerves_hub_web`

PR #194, opened 2024-05-24
retry and fetch an updated firmware url if the first try 401s

joshk:
this is a fix for #47 

if the download returns a 401 it will retry once and then exit

I have also trapped `Downloader` exits and nil'd `UpdateManager` state where appropriate


test/nerves_hub_link/update_manager_test.exs
```diff
@@ -1,12 +1,16 @@
...
   alias NervesHubLink.{FwupConfig, UpdateManager}
   alias NervesHubLink.Message.{FirmwareMetadata, UpdateInfo}
   alias NervesHubLink.Support.FWUPStreamPlug
 
+  import Mock
+
   describe "fwup stream" do
     setup do
-      port = 6000
+      port = Enum.random(6000..6999)
```
lawik, line 12: Random here but specific in another test?

test/nerves_hub_link/update_manager_test.exs
```diff
@@ -87,6 +91,51 @@ defmodule NervesHubLink.UpdateManagerTest do
     end
   end
 
+  describe "401 retry" do
+    setup do
+      port = 6543
```
lawik, line 96: fixed port number here?

lawik (commented):
Some notes on port numbering in tests but no major weirdness :)

test/nerves_hub_link/update_manager_test.exs
joshk: There was an error in a Circle run where a test failed due to the port being in use, which is odd as the test was not async, but I added this as a way to make sure it didn't happen again.

test/support/http_unauthorized_error_plug.ex
```diff
@@ -0,0 +1,18 @@
...
+  @behaviour Plug
+
+  import Plug.Conn
+
+  @impl Plug
+  def init(options), do: options
+
+  @impl Plug
+  def call(conn, opts) do
+    send(opts[:report_pid], {:request_error})
```
jjcarstens, line 13:
```suggestion
    send(opts[:report_pid], :request_error)
```
Prob doesn't need to be a tuple?

test/support/http_unauthorized_error_plug.ex
```diff
@@ -0,0 +1,18 @@
...
+
+    conn
+    |> send_resp(401, "Unauthorized")
```
jjcarstens, line 16:
```suggestion
    send_resp(conn, 401, "Unauthorized")
```

test/nerves_hub_link/update_manager_test.exs
```diff
@@ -87,6 +91,51 @@ defmodule NervesHubLink.UpdateManagerTest do
...
+      {:ok, manager} = UpdateManager.start_link(fwup_config)
+
+      with_mock NervesHubLink.Socket, check_update_available: fn -> update_payload end do
+        assert UpdateManager.apply_update(manager, update_payload, []) == {:updating, 0}
+
+        assert_receive {:fwup, {:progress, 0}}, 1_000
+        assert_receive {:request_error}, 1_000
+        assert_receive {:request_error}, 1_000
+
+        Process.sleep(300)
```
jjcarstens, line 132: Is there any way around having a hard sleep in the test?

test/nerves_hub_link/update_manager_test.exs
```diff
@@ -87,6 +91,51 @@ defmodule NervesHubLink.UpdateManagerTest do
     end
   end
 
+  describe "401 retry" do
+    setup do
+      port = 6543
+      devpath = "/tmp/fwup_output"
```
jjcarstens, line 97: Great use case for a tmp dir which should be unique to the test and you won't have to worry about cleaning up

https://hexdocs.pm/ex_unit/ExUnit.Case.html#module-tmp-dir

lib/nerves_hub_link/update_manager.ex
```diff
@@ -61,6 +67,14 @@ defmodule NervesHubLink.UpdateManager do
     GenServer.call(manager, :status)
   end
 
+  @doc """
+  Returns the current status of the update manager
```
jjcarstens, line 71:
```suggestion
  Returns the previous update status of the update manager
```

This says current status, but the function is previous update?

lib/nerves_hub_link/update_manager.ex
```diff
@@ -145,22 +163,55 @@ defmodule NervesHubLink.UpdateManager do
...
+        {:download, {:error, %Mint.HTTPError{reason: {:http_error, 401}}}, fwup_public_keys},
+        state
+      ) do
+    Logger.error(
+      "[NervesHubLink] Firmware download URL signature may have expired. Requesting new URL and retrying."
+    )
+
+    GenServer.stop(Fwup.Stream)
+
+    update_info = Socket.check_update_available()
```
jjcarstens, line 190: It's prob worth making this async. With this, you have a hard limit of 5 seconds before the genserver call times out. If the request to the server and resolution of the update takes 5.1 seconds, then this all fails

Switching to async means you fire a message to the server, it can take its time getting back, and you know how to respond to the next update message when received.

If we're really worried about it, maybe we have an UpdateRetry that fails after a few minutes to indicate the server never told it a new URL? (Or just a log message)

lib/nerves_hub_link/update_manager.ex
joshk: Thanks for spotting this!

test/nerves_hub_link/update_manager_test.exs
joshk: Ohhhh, nice, I've switched to this, and updated other tests to use it too

test/support/http_unauthorized_error_plug.ex
joshk: true, I ran into an issue where I thought I needed the curly brackets, but I resolved the issue.

thanks!

test/nerves_hub_link/update_manager_test.exs
joshk: hmmmm, I added it because the final assert needed time for the `previous_update` status to be set, but I've run some tests over and over without it and it seems to be fine. I've now removed it, and if issues arise then I might need some help.

lib/nerves_hub_link/update_manager.ex
joshk: I agree that async would be better.

It will take some reworking, including in the Hub Web code.

joshk:
@jjcarstens I'm wondering if maybe the better way to approach this issue is to have `UpdateManager` report back to the `Socket` on the status of the update (rescheduled, ignored, download error), and then Hub can decide if it sends the update again.

Thoughts?

lawik:
For time blog out of whack, one solution is to have the device include what they think the time is in join.

Then the server can tell whether the client is confused or has an expired thing. It can even pass down the current time. This could be an optional mechanism for getting the time (trust the server) which is helpful for a Pi for example. But mostly it lets one tell the difference between bad auth and being out of sync.

jjcarstens:
A noble thought @lawik, but I don't think that will work. Time is a required component _of the connect_ for a couple reasons:
* Device certs - the SSL stack on device may not even get to a connect attempt because it will check the server cert first and believe it is expired when it compares its terribly off device time
* Shared secret - Time is a component of the HMAC to assert validity. To detect the time is off and say "o, that's okay. Here's the time you should use" would be an anti-pattern negating the HMAC a bit

However, the middle ground is a decent idea to just include the `date` header in the rejection always. Though I am unsure if we could get that early enough in the SSL stack. Though Frank is currently working on a time server solution that could be used for conditions like this too which might be the better play to keep the time concern separate

lawik:
Shared secret has this issue. The SSL is not failing, it gets a 401 and if we knew what it thinks the time is we can detect it and give a better "why is NervesHubLink not up"-experience.

My PIs have this and log a ton of 401s on every restart. My guess is that they generate times at UTC until NervesTime syncs. Or something. Time is not 1977-01-01 :) SSL works it seems

joshk:
@jjcarstens I've reworked the logic of this feature to now send updates from the `UpdateManager` to the pid that called it. These updates are then sent to Hub, which can then update the registry and decide if there are further actions to take, like sending a new update to the device.

I think this is a better setup, and there are options to improve it further.

lib/nerves_hub_link/update_manager.ex
```diff
@@ -145,26 +163,48 @@ defmodule NervesHubLink.UpdateManager do
...
+    {:noreply, %State{state | status: :idle}}
+  end
+
+  def handle_info(
+        {:download, {:error, %Mint.HTTPError{reason: {:http_error, 401}}}, _fwup_public_keys,
+         reporting_pid},
+        state
+      ) do
+    Logger.error("[NervesHubLink] Firmware download error: 401")
+    send(reporting_pid, {:update_manager, {:error, :download_unauthorized}})
```
jjcarstens, line 178: I don't think results can be a tuple since this message eventually gets pushed up the socket and would not be JSON encodable

joshk: I've fixed this by formatting the possible response types when sending to Hub in `socket.ex`

lib/nerves_hub_link/socket.ex
```diff
@@ -443,6 +443,19 @@ defmodule NervesHubLink.Socket do
     {:noreply, stop_iex(socket)}
   end
 
+  def handle_info({:update_manager, result}, socket) do
+    formatted =
+      if is_atom(result) do
+        %{result: result}
+      else
+        %{result: elem(result, 0), reason: elem(result, 1)}
```
jjcarstens, line 451: If we're not doing anything with the `{:error` portion of the tuple, why have it? We could just simplify and make it an atom return

jjcarstens: That or change the success condition to be `{:ok, result}` and we always return the same structure of `%{result: :ok | :error, reason: reason}`

joshk: there are six events that can be sent from the `UpdateManager` to `Socket` are, the update ...

- started
- was ignored
- was rescheduled
- completed successfully
- failed due to a 401 error (`download_unauthorized`)
- failed due to a non fatal download error (`non_fatal`)

I had modelled this as ...
- `:started`
- `:ignored`
- `{:rescheduled, ms, rescheduled_to(ms)}`
- `:completed`
- `{:error, :download_unauthorized}`
- and `{:error, :non_fatal}`

I could change `:completed` to `{:ok, :completed}` to match the structure of the error cases, although the `:ok` is kinda superfluous information. It would also mean that `:ignored` and `:started` would need to change.

In Hub each of these cases would update registry in slightly different ways.

Thoughts?

I also realized that I hadn't accounted for `{:rescheduled, ms, rescheduled_to(ms)}`

PR #195, opened 2024-05-24
check time synchronization before socket connection

### LINKED ISSUE #191: Link fails with 403 repeatedly until it eventually recovers
```
06:48:16.099 [warn] [NervesHubLink] error: {:error, {:upgrade_failure, %{reason: %Mint.WebSocket.UpgradeFailureError{status_code: 403, headers: [{"date", "Fri, 17 May 2024 06:48:15 GMT"}, {"content-length", "0"}, {"vary", "accept-encoding"}, {"cache-control", "max-age=0, private, must-revalidate"}]}, status_code: 403, resp_headers: [{"date", "Fri, 17 May 2024 06:48:15 GMT"}, {"content-length", "0"}, {"vary", "accept-encoding"}, {"cache-control", "max-age=0, private, must-revalidate"}]}}}

06:48:16.099 [info]     :alarm_handler: {:set,
 {NervesHubLink.Disconnected,
  [
    reason: {:error,
     {:upgrade_failure,
      %{
        reason: %Mint.WebSocket.UpgradeFailureError{
          status_code: 403,
          headers: [
            {"date", "Fri, 17 May 2024 06:48:15 GMT"},
            {"content-length", "0"},
            {"vary", "accept-encoding"},
            {"cache-control", "max-age=0, private, must-revalidate"}
          ]
        },
        status_code: 403,
        resp_headers: [
          {"date", "Fri, 17 May 2024 06:48:15 GMT"},
          {"content-length", "0"},
          {"vary", "accept-encoding"},
          {"cache-control", "max-age=0, private, must-revalidate"}
        ]
      }}}
  ]}}

06:48:16.108 [warn] [NervesHubLink.Socket] Unhandled handle_info: {:EXIT, #Port<0.100>, :normal}

06:48:32.561 [info] disconnect --------

06:48:32.562 [warn] [NervesHubLink] error: {:error, {:upgrade_failure, %{reason: %Mint.WebSocket.UpgradeFailureError{status_code: 403, headers: [{"date", "Fri, 17 May 2024 06:48:31 GMT"}, {"content-length", "0"}, {"vary", "accept-encoding"}, {"cache-control", "max-age=0, private, must-revalidate"}]}, status_code: 403, resp_headers: [{"date", "Fri, 17 May 2024 06:48:31 GMT"}, {"content-length", "0"}, {"vary", "accept-encoding"}, {"cache-control", "max-age=0, private, must-revalidate"}]}}}

06:48:32.563 [info]     :alarm_handler: {:set,
 {NervesHubLink.Disconnected,
  [
    reason: {:error,
     {:upgrade_failure,
      %{
        reason: %Mint.WebSocket.UpgradeFailureError{
          status_code: 403,
          headers: [
            {"date", "Fri, 17 May 2024 06:48:31 GMT"},
            {"content-length", "0"},
            {"vary", "accept-encoding"},
            {"cache-control", "max-age=0, private, must-revalidate"}
          ]
        },
        status_code: 403,
        resp_headers: [
          {"date", "Fri, 17 May 2024 06:48:31 GMT"},
          {"content-length", "0"},
          {"vary", "accept-encoding"},
          {"cache-control", "max-age=0, private, must-revalidate"}
        ]
      }}}
  ]}}

06:48:32.572 [warn] [NervesHubLink.Socket] Unhandled handle_info: {:EXIT, #Port<0.105>, :normal}
```

This might be the setup of NervesCloud @joshk, or it could be normal thing.

Is there a token or something written to the data partition or something. It seems to happen after I've done a firmware update, it starts up, gets networking but fails to auth?

joshk:
Similar to the recent addition of checking if the network is available before connecting the socket, this also checks if the time is synchronized. 

Additionally, you can opt-out by using:

```elixir
config :nerves_hub_link,
  wait_for_time_sync: false,
```

which is useful for tests or when running your code on your host machine (eg. Mac) where NTP isn't available and `NervesTime` will always report that time isn't synchronized.

fixes #191

### DIFF
--- a/lib/nerves_hub_link/socket.ex
+++ b/lib/nerves_hub_link/socket.ex
@@ -90,12 +90,8 @@ defmodule NervesHubLink.Socket do
       |> assign(connected_at: nil)
       |> assign(joined_at: nil)
 
-    if config.connect_wait_for_network do
-      schedule_network_availability_check()
-      {:ok, socket}
-    else
-      {:ok, socket, {:continue, :connect}}
-    end
+    schedule_preconnect_checks()
+    {:ok, socket}
   end
 
   @impl Slipstream
@@ -388,15 +384,25 @@ defmodule NervesHubLink.Socket do
   end
 
   @impl Slipstream
-  def handle_info(:connect_check_network_availability, socket) do
-    case :inet.gethostbyname(to_charlist(socket.assigns.config.device_api_host)) do
-      {:ok, _} ->
+  def handle_info(:preconnect_checks, socket) do
+    network? = network_connected?(socket)
+    time? = time_synchronized?(socket)
+
+    cond do
+      network? && time? ->
         {:noreply, socket, {:continue, :connect}}
 
-      _ ->
-        Logger.info("[NervesHubLink] waiting for network to become available")
-        schedule_network_availability_check(2_000)
-        {:noreply, socket}
+      !network? ->
+        log_and_reschedule_checks(
+          "[NervesHubLink] waiting for network to become available",
+          socket
+        )
+
+      !time? ->
+        log_and_reschedule_checks(
+          "[NervesHubLink] waiting for time to sync",
+          socket
+        )
     end
   end
 
@@ -447,6 +453,33 @@ defmodule NervesHubLink.Socket do
     {:noreply, socket}
   end
 
+  defp network_connected?(%{assigns: %{config: config}}) do
+    if config.wait_for_network do
+      host = to_charlist(config.device_api_host)
+
+      case :inet.gethostbyname(host) do
+        {:ok, _} -> true
+        _ -> false
+      end
+    else
+      true
+    end
+  end
+
+  defp time_synchronized?(%{assigns: %{config: config}}) do
+    if config.wait_for_time_sync do
+      NervesTime.synchronized?()
+    else
+      true
+    end
+  end
+
+  defp log_and_reschedule_checks(msg, socket) do
+    Logger.info(msg)
+    schedule_preconnect_checks(2_000)
+    {:noreply, socket}
+  end
+
   @impl Slipstream
   def handle_topic_close(topic, reason, socket) when reason != :left do
     if topic == @device_topic do
@@ -485,8 +518,8 @@ defmodule NervesHubLink.Socket do
     disconnect(socket)
   end
 
-  defp schedule_network_availability_check(delay \\ 100) do
-    Process.send_after(self(), :connect_check_network_availability, delay)
+  defp schedule_preconnect_checks(delay \\ 100) do
+    Process.send_after(self(), :preconnect_checks, delay)
   end
 
   defp handle_join_reply(%{"firmware_url" => url} = update, socket) when is_binary(url) do
--- a/lib/nerves_hub_link/configurator.ex
+++ b/lib/nerves_hub_link/configurator.ex
@@ -9,7 +9,6 @@ defmodule NervesHubLink.Configurator do
   defmodule Config do
     defstruct archive_public_keys: [],
               connect: true,
-              connect_wait_for_network: true,
               data_path: "/data/nerves-hub",
               device_api_host: nil,
               device_api_port: 443,
@@ -25,12 +24,13 @@ defmodule NervesHubLink.Configurator do
               request_fwup_public_keys: false,
               shared_secret: [],
               socket: [],
-              ssl: []
+              ssl: [],
+              wait_for_network: true,
+              wait_for_time_sync: true
 
     @type t() :: %__MODULE__{
             archive_public_keys: [binary()],
             connect: boolean(),
-            connect_wait_for_network: boolean(),
             data_path: Path.t(),
             device_api_host: String.t(),
             device_api_port: String.t(),
@@ -46,7 +46,9 @@ defmodule NervesHubLink.Configurator do
             request_fwup_public_keys: boolean(),
             shared_secret: [product_key: String.t(), product_secret: String.t()],
             socket: any(),
-            ssl: [:ssl.tls_client_option()]
+            ssl: [:ssl.tls_client_option()],
+            wait_for_network: boolean(),
+            wait_for_time_sync: boolean()
           }
   end
 


lawik (commented):
I like it :)

fhunleth:
I feel like there should be a warning. If you were to put this into production and you don't have control over where the device is installed, it might prevent the device from connecting. NTP is sometimes blocked by routers. I don't know if it's a blanket UDP ban or whether some admin really doesn't like NTP, but we see it happen in our installs. NTP synchronization also takes minutes, so if something causes the device to reboot soon after boot, it might also make it trickier to recover the device.

joshk:
That is a good point, your time can be in sync without NTP.

I do wonder if having a HTTP based NTP adapter for `NervesTime` might be a slight better way to go for the case you mention above. As a devices clock drifts, it will cause issues with SSL.

For this PR, would the warning be during socket connection?

lawik:
Good call.

So the problem seems to have been that the client tries to connect with a clock that is out of whack and fails to do the crypto dance. I get this on every restart as noted in #191.

It isn't harmful to fail that way until it succeeds but it is kind of weird.

Would time NervesTime.synchronized? give false even if an RTC has kept it accurate? Then synchronization is the wrong thing to look at.

lib/nerves_hub_link/socket.ex
```diff
@@ -485,8 +518,8 @@ defmodule NervesHubLink.Socket do
     disconnect(socket)
   end
 
-  defp schedule_network_availability_check(delay \\ 100) do
-    Process.send_after(self(), :connect_check_network_availability, delay)
+  defp schedule_network_and_time_check(delay \\ 100) do
```
jjcarstens, line 521:
```suggestion
  defp schedule_preconnect_checks(delay \\ 100) do
```
Maybe we keep this slightly generic to give an entry point running all the checks we need? (At your rate, there might be more coming 🤣)

lib/nerves_hub_link/socket.ex
```diff
@@ -388,15 +384,25 @@ defmodule NervesHubLink.Socket do
...
 
-      _ ->
-        Logger.info("[NervesHubLink] waiting for network to become available")
-        schedule_network_availability_check(2_000)
-        {:noreply, socket}
+      !network_connected?(socket) && !time_synchronized?(socket) ->
+        network_check_with_message(
+          "[NervesHubLink] waiting for network to become available and time to sync",
+          socket
+        )
```
jjcarstens, line 396: Do we need to be this verbose? As in reporting every possible condition vs just the first of time or connect problem we detect?

lib/nerves_hub_link/socket.ex
```diff
@@ -447,6 +453,33 @@ defmodule NervesHubLink.Socket do
...
+
+  defp time_synchronized?(%{assigns: %{config: config}}) do
+    if config.wait_for_time_sync do
+      NervesTime.synchronized?()
+    else
+      true
+    end
+  end
+
+  defp network_check_with_message(msg, socket) do
```
jjcarstens, line 477:
```suggestion
  defp log_and_reschedule_checks(msg, socket) do
```
What do you think of a more purpose built function name?

lib/nerves_hub_link/socket.ex
joshk: hahaha, I like your naming, and I promise not to add more (for a while)

lib/nerves_hub_link/socket.ex
joshk: your naming is much better

lib/nerves_hub_link/socket.ex
joshk: Thats a good point, and I agree, and yet I flip flop.

I think you are right though. I'll change this around.

lawik:
I wonder if it is better to just try and figure out if the time is out of whack based on the request rather than not making it before time sync. We can handle the outcome if we know that it is the problem.

joshk:
what is the error you get when the time isn't synchronized yet?

lawik:
403, see https://github.com/nerves-hub/nerves_hub_link/issues/191

joshk:
Thats right. 403 doesn't really tell us anything, that is a generic error, or as Mozilla tells me...

> The HTTP 403 Forbidden response status code indicates that the server understands the request but refuses to authorize it.

lawik:
Issue has more details. It fails to establish the websocket and I bekieve the 403 is only triggered by an invalid signature. The signature is invalid because it signs a couple of hours off before NervesTime syncs up.

we might not want to wait for NervesTime sync. But if we don't we should have a warning that this could be due to time kerfuffery so people don't scratch their heads for two minutes and then change the config for their perfectly correct firmware.

lawik:
The server could pass its perceived time as part of the 403 in a header or something. Would let us check for this case. Maybe do some backoff?

jjcarstens (approved):

fhunleth:
I'm really having trouble with this one. I don't think that I'd ever enable it or recommend anyone enable it. The reason is that it assumes failure without trying. Maybe the time is wrong, but maybe it will never be fixed because NTP is down or blocked. If the devices never try, then the only option is to replace them in the field or hope that there's another backend connection that doesn't do this. On the other hand, if the devices try, then this can be addressed on the backend by deploying a cert with a better validity range. 

I agree with the sentiment of just fixing the error message. How about if the SSL connection fails, then check if the time has been synchronized. Same logic, but be optimistic and try rather than assume failure if unsynchronized.

lawik (changes requested):
See Frank's note. I don't think we should push this time stuff through.

I think we can make the dev aware of the 403 problem and why it might happen and they should have less of that when the ping endpoints exist.

For more serious cases this might be causing more trouble than worth.

joshk:
Thanks for the feedback @fhunleth 

I'll look into the changes you mentioned.

One thing to note is that these 403 errors due to time synchronization push the Socket retry backoff to hit its max limit. This is a similar issue we saw several months ago related to the Shared Secrets work, and likely means there is some work to be done there (like resetting the backoff after a successful connection)

lawik:
We've done work to mitigate this in another issue where we were using `System.system_time` instead of `System.os_time` and the system time would take a long time to update while OS time was more reasonable.

I think we should change the warning but this thread can be closed.

PR #196, opened 2024-05-24
request a port as needed

joshk:
its likely overkill, but it does mean we don't need to remember which port number is being defined in which test

### DIFF
--- a/test/nerves_hub_link/downloader_test.exs
+++ b/test/nerves_hub_link/downloader_test.exs
@@ -6,6 +6,7 @@ defmodule NervesHubLink.DownloaderTest do
     IdleTimeoutPlug,
     RangeRequestPlug,
     RedirectPlug,
+    Utils,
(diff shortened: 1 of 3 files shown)


lawik:
I like it. But clearly inefficient :P
ETS has atomic counters.

But confirming that I ran across this thread: https://elixirforum.com/t/is-ets-update-counter-efficient/20359/6

And it referenced Erlang :counters:
https://www.erlang.org/doc/apps/erts/counters.html

That should be the Performant Solution ™️ 

Then it would be overkill.

joshk:
@jjcarstens suggested in Slack to use `System.unique_integer([:positive, :monotonic])` instead, a much simpler approach.

lawik:
I considered that but didn't want to think about how to get it within the range of port numbers.

jjcarstens (approved):

PR #201, opened 2024-06-13
Health checks

lawik:


lib/nerves_hub_link/health_check.ex
```diff
@@ -0,0 +1,49 @@
+defmodule NervesHubLink.HealthCheck do
```
joshk, line 1: We can skip this file and add the smarts to the `Client` module instead.

lib/nerves_hub_link/health_check/default_report.ex
```diff
@@ -0,0 +1,122 @@
...
+        # TODO: What else?
+        %{
+            "arch" => KV.get_active("nerves_fw_architecture"),
+            "platform" => KV.get_active("nerves_fw_platform"),
+            "product" => KV.get_active("nerves_fw_product"),
+            "version" => KV.get_active("nerves_fw_version"),
+            "uuid" => KV.get_active("nerves_fw_uuid"),
+        }
+        |> Map.merge(metadata_from_config())
+    end
```
joshk, line 24: I'd skip the `device_id`, `timestamp` and the default `metadata` information too.

All this data is known by Hub when the connection is established.

lib/nerves_hub_link/health_check/default_report.ex
```diff
@@ -0,0 +1,122 @@
...
+                    {:ok, [1min, 5min, 15min]} ->
+                        Map.merge(report, %{load_1min: 1min, load_5min: 5min, load_15min: 15min})
+                    end
+    
+        Map.merge(report, metrics_from_config())
+    end
+
+    def peripherals do
+        peripherals =
+            for ifname <- VintageNet.all_interfaces(), into: %{}, do
```
joshk, line 70: is it guaranteed that `VintageNet` is loaded and used?

joshk (commented):
I think the next bit of code needed is in `Socket` which would reply to this information when asked for it.

lib/nerves_hub_link/health_check.ex
lawik: Will look into that :)

lawik:
> I think the next bit of code needed is in `Socket` which would reply to this information when asked for it.

I've added this stuff now. I think this implementation is "complete" for now. If completely untested :D I could add the other end too unless you feel extra keen @joshk

lib/nerves_hub_link/client/default.ex
```diff
@@ -77,4 +79,25 @@ defmodule NervesHubLink.Client.Default do
...
+    config = Application.get_env(:nerves_hub_link, :health)
+
+    if config == false do
+      # No check at all, disabled
+      nil
+    else
+      report = config[:report] || NervesHubLink.HealthCheck.DefaultReport
+
+      DeviceStatus.new(
+        device_id: report.device_id(),
```
joshk, line 94: was leaving this in intentional?

lib/nerves_hub_link/health_check.ex
```diff
@@ -0,0 +1,27 @@
...
```
joshk, line 1: whats the purpose of this file?

joshk (commented):
some small questions

is it worth adding some light tests for this?

lawik:
We should add tests. This is still a draft :)

lib/nerves_hub_link/client/default.ex
lawik: Nope, will fix :)

lib/nerves_hub_link/health_check.ex
lawik: Will move the doc of the config stuff. Didn't want to remove them until I had put them somewhere.

lib/nerves_hub_link/health_check/report.ex
```diff
@@ -1,7 +1,7 @@
 defmodule NervesHubLink.HealthCheck.Report do
   alias NervesHubLink.Message.DeviceStatus.Peripheral
 
-  @callback timestamp() :: DateTime.t()
+  @callback timestamp() :: String.t()
```
lawik, line 4: This was intentional. They serialize well and this means we avoid sending bad timestamps.

lib/nerves_hub_link/health_check/default_report.ex
```diff
@@ -0,0 +1,126 @@
...
+      %{cpu_temp: temp}
+    else
+      _ -> %{}
+    end
+  end
+
+  defp load_averages do
+    with {:ok, data_str} <- File.read("/proc/loadavg"),
+         [min1, min5, min15, _, _] <- String.split(data_str, " ") do
+      %{load_1min: min1, load_5min: min5, load_15min: min15}
```
lawik, line 109: Why remove the load parsing into numbers? Metrics are supposed to be numbers, type-wise?

lib/nerves_hub_link/health_check/report.ex
joshk: Dialyzer was exploding in CI due to this. You can switch it back, I just went for the quick fix.

lib/nerves_hub_link/health_check/default_report.ex
joshk: you were change the percentages into just 0 or 1, and then I thought about sending up floats, but after talking with Jon, I opted for the parsing of this on hub, as hub will need to do normalization.

lib/nerves_hub_link/health_check/report.ex
lawik: What's wrong with it according to dialyzer?

lib/nerves_hub_link/health_check/default_report.ex
joshk: I'm not tied to strings, we can send floats. We will likely want to also do normalization in hub too.

lawik: Ah, that's a goof on my part to use Integers. The typing is `number()` specifically to support float and int which also works with JSON.

lib/nerves_hub_link.ex
```diff
@@ -55,4 +57,20 @@ defmodule NervesHubLink do
   """
   @spec send_file(Path.t()) :: :ok | {:error, :too_large | File.posix()}
   defdelegate send_file(file_path), to: Socket
+
+  @doc """
+  Send a health check to NervesHub
+  """
+  @spec send_health_check() :: Message.DeviceStatus.t()
+  def send_health_check() do
```
joshk, line 65: the only thing that might be handy is catching errors here (since people can create their own reports) and send up an error message to hub, something about how there was an error generating the report?

lawik: Interesting, yeah, somewhere in the chain, that should be a thing.

joshk: That could work, although a little funny/odd that the service that generates report that uses alarms also generates alarms

joshk:
I think we might want to bump https://github.com/nerves-hub/nerves_hub_link/blob/main/lib/nerves_hub_link/configurator.ex#L6 when this is merged in, and we can use this information in hub to decide to display information or a message to upgrade.

lawik:
> I think we might want to bump https://github.com/nerves-hub/nerves_hub_link/blob/main/lib/nerves_hub_link/configurator.ex#L6 when this is merged in, and we can use this information in hub to decide to display information or a message to upgrade.

But I think we've just added stuff? Or did we change any of the API surface already in use?

joshk:
we added support for requesting health data. by matching this to the api version we can display in the UI that Link needs to be updated for this information to become available.

PR #202, opened 2024-06-19
Apply backpressure to downloader on slow updates

fhunleth:
It's possible for the firmware download to be faster than fwup can write
to eMMC or MicroSD. This causes the message queue to the update manager
to balloon as data starts queuing up in route to fwup. On a device
without much extra memory, this can be a cause of the following error:

```
04:14:26.957 [warn] __vm_enough_memory: pid: 87, comm: erts_sched_1, not enough memory for the allocation
```

This commit makes the asynchronous downloader updates synchronous so
fwup can push back when the writes are time consuming. The backpressure
results in the TCP receive buffers filling and eventually the download
slowing.

### DIFF
--- a/lib/nerves_hub_link/update_manager.ex
+++ b/lib/nerves_hub_link/update_manager.ex
@@ -69,6 +69,19 @@ defmodule NervesHubLink.UpdateManager do
     GenServer.call(manager, :currently_downloading_uuid)
   end
 
+  # Private API for reporting download progress. This wraps a GenServer.call so
+  # that it can apply backpressure to the downloader if applying the update is
+  # slow.
+  defp report_download(manager, message) do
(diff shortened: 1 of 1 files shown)


fhunleth:
I have this as draft since I've only run the unit tests. @ringlej and @abelino might be able to test it on a device that runs into the out of memory error.

fhunleth:
See https://github.com/nerves-hub/nerves_hub_link_common/pull/35 for what was actually tested. @ringlej is using the `maint_v1` branch.

jjcarstens (approved):

oestrich:
There's something not quite right with this on a real device. I need to keep poking at it before we merge

oestrich:
The issue was something else and this seems to have the intended effect of not being noticeable on real devices :shipit:

PR #204, opened 2024-06-24
Remove the `[] =` for fwup_public_keys

oestrich:
This caused a pattern match error when public key(s) were present because this was matching against an empty list.

### DIFF
--- a/lib/nerves_hub_link/update_manager.ex
+++ b/lib/nerves_hub_link/update_manager.ex
@@ -213,7 +213,7 @@ defmodule NervesHubLink.UpdateManager do
   end
 
   @spec start_fwup_stream(UpdateInfo.t(), [binary()], State.t()) :: State.t()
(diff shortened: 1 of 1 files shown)


fhunleth:
How did this ever work? Is it recent?

oestrich:
It’s from maybe a month ago but we haven’t had a release from even longer. Guess no one ran it when we merged it? 🤷

fhunleth:
Which commit/PR?

oestrich:
This was the pr https://github.com/nerves-hub/nerves_hub_link/pull/187

fhunleth:
Could you revise the commit message so that it's (or revise if I got it wrong):

```
Fix regression that prevents using fwup public keys

Commit 3275db2e added an empty list pattern match that would
fail when a key was present. Fix issue by removing the match.
```

Change looks good.

joshk:
Thanks for catching this @oestrich

PR #205, opened 2024-06-24
heartbeat_interval_msec is a top level config value

oestrich:


fhunleth (approved):

joshk:
Thanks for catching this @oestrich 

@fhunleth @oestrich , should we consider moving `reconnect_after_msec` to the top level as well to match?

fhunleth:
@joshk I'm not opposed, but it would be a backwards incompatible change. Perhaps next consider for next release? I'm anxious to get the back pressure fix out.

joshk:
Totally understand.

I’d suggest we move the setting under socket to match, but I don’t want to
hold up the release.

PR #206, opened 2024-06-24
v2.4.0 release

oestrich:
* Added
  * Configure heartbeat msec (#197)
  * Apply backpressure to downloader on slow updates (#202)
* Updated
  * Fix no match error when uploading files (#190)
  * Make shared secret auth the default (#176)
  * Archive public keys can default to fwup public keys if not configured (#187)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## unreleased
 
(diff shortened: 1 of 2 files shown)


fhunleth (approved):

joshk (approved):
thanks for putting this together @oestrich

PR #207, opened 2024-06-25
Run a NervesHub script on device

oestrich:
Evaluate a script locally on device from NervesHub instead of relying on the console channel. Bumps the api version so NervesHub is aware this is possible on the connecting device.

### DIFF
--- a/lib/nerves_hub_link/script.ex
+++ b/lib/nerves_hub_link/script.ex
@@ -1,19 +1,32 @@
 defmodule NervesHubLink.Script do
-  # Inspired from ExUnit.CaptureIO
-  # https://github.com/elixir-lang/elixir/blob/main/lib/ex_unit/lib/ex_unit/capture_io.ex
+  use GenServer
 
   @doc """
   Run a script from NervesHub and capture its output
   """
-  def capture(text) do
-    original_gl = Process.group_leader()
+  def capture(text, ref) do
+    _ = GenServer.start_link(__MODULE__, {self(), text, ref})
+
+    :ok
+  end
+
+  @impl GenServer
+  def init({pid, text, ref}) do
+    state = %{pid: pid, text: text, ref: ref}
+    {:ok, state, {:continue, :capture}}
+  end
+
+  @impl GenServer
+  def handle_continue(:capture, state) do
+    # Inspired from ExUnit.CaptureIO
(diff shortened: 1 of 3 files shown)


fhunleth (approved):

jjcarstens (commented):
Chatted with Eric on the side. I have a couple concerns here:
* The script evaluation is blocking in `handle_message` on the Socket. If a bad script comes, it would cause the socket to stop responding to messages
* The current `Script.capture/1` is altering the group leader of the caller, which is also the Socket in this point.

Based on those, we probably want this to run in it's own process so it only affects a transient group leader and doesn't block the socket

lib/nerves_hub_link/script.ex
```diff
@@ -0,0 +1,46 @@
+defmodule NervesHubLink.Script do
+  use GenServer
+
+  @doc """
+  Run a script from NervesHub and capture its output
+  """
+  def capture(text, ref) do
+    _ = GenServer.start(__MODULE__, {self(), text, ref})
```
fhunleth, line 8: Shouldn't this be `start_link`'d so that if the process calling `capture/2`, the code doesn't continue to run? I also wonder if there should be a limit to the number of scripts being run simultaneously.

oestrich: No, I don’t think so at least. Since we didn’t want the script to kill the socket we don’t want to link it during capture

fhunleth: If the script crashes, an error should be returned to the server. If the connection crashes for other reasons, then is there any reason for the script to run since it can't return results to the server any more?

oestrich: I’m going to assume I misunderstood what Jon wanted since what you described is how it worked before

fhunleth: Yeah, nothing can be easy, can it? I took Jon's comment

It feels like this may be a temporary fix to an immediately pressing issue, and we'll be revisiting it soon. In case you haven't seen other threads, there's #220 which is discussing an extension system. This would make it easier to add non-firmware update features without worrying about the possible ramifications to the core functionality.

I do think that we should return errors when things go wrong to make issues easier to debug. I read Jon's comment as don't block the main connection process since it would block processing of firmware update requests.

fhunleth: It seems like we can wait for async feedback over the next day, and then decide what's most practical to do for #221 tomorrow.

PR #209, opened 2024-06-28
Add option to override serial_number for running on host

lawik:


lawik:
Need to test this with a patched up nerves_time because they are fighting.

lib/nerves_hub_link/application.ex
```diff
@@ -26,6 +26,8 @@ defmodule NervesHubLink.Application do
 
   defp children(%{connect: false}, _fwup_config), do: []
 
+  defp children(%{firmware_updates: false} = config, _fwup_config), do: [{Socket, config}]
```
jjcarstens, line 29: If we do this, can we do a big fat warning that yells it is disabled with the instruction for turning it back on? (maybe at compile time)

I worry about too many of these developer conveniences making it too easy to prevent a real device from connecting and updating.

Alternatively, if you still want to see the updates attempt, you could also try changing `fwup_devpath: "/tmp/some_bogus_devpath"` - Fwup will write to that temp file and you get all the progress

lawik: Oh, happy to try that for sure. I believe the explosions I was mitigating were around the fwup_devpath. I think I misread that as fwup_path and just noped out of it.

lawik: I get an initial thing where the socket crashes on join and then joins clean. I think that's Nerves.Runtime not having the value on the first attempt. But not sure. Good enough to be useful though and we can harden nerves_hub_web to be more graceful potentially.

lawik:
Now this only offers the serial number override.

oestrich:
You can also override the serial number by the host config for the nerves runtime kv. I think you need to set this anyways for nerves hub link to function properly on a host? 

This is what I have in my dev app for nerves hub 

```
config :nerves_runtime,
  kv_backend:
    {Nerves.Runtime.KVBackend.InMemory,
     contents: %{
       "nerves_fw_active" => "a",
       "a.nerves_fw_uuid" => "6d5566a3-0b09-5ed2-841d-0223c77ce702",
       "a.nerves_fw_product" => "forgery",
       "a.nerves_fw_architecture" => "arm",
       "a.nerves_fw_version" => "0.5.1",
       "a.nerves_fw_platform" => "rpi0",
       "a.nerves_fw_misc" => "extra comments",
       "a.nerves_fw_description" => "test firmware",
       "nerves_fw_devpath" => "/tmp/fwup_bogus_path",
       "nerves_serial_number" => "forgery"
     }}
```

lawik:
I did set that. But nerves_hub_link talks directly to Nerves.Runtime.serial_number which will directly ask board id and my "device" gets the name "unconfigured". Hence, this :)

oestrich:
Ah, this is for the shared secret way of connecting. That explains why I didn’t need to set it to get connected properly since I’ve got a cert

lawik:
That explains that :)

lawik:
Hold on, is there some other call we could make that would vrab it from KV and is will KV pick up from boardid?

jjcarstens:
The way I get around this is I make a small script for a serial number that I set as the boardid file

```sh
#!/bin/sh

# pick your favorite serial number by default
# or set it on run with SERIAL=abc123 iex -S mix
serial=${SERIAL:-default_1234}

echo -n "$serial"
```

```elixir
# some config.exs for host

config :nerves_runtime, boardid_path: /path/to/host/boardid
```

I'm not sure this needs to go in a library but prob a tip as it applies to all the libs you want to test on host (which I'm a huge advocate for)

lawik:
Josh had a thing for it already:

```
config :nerves_hub_link,
  device_api_host: "devices.nervescloud.com",
  shared_secret: [
    product_key: System.get_env("NERVES_HUB_KEY"),
    product_secret: System.get_env("NERVES_HUB_SECRET"),
    identifier: System.fetch_env!("NERVES_SERIAL_NUMBER") # THIS
  ]
```

PR #214, opened 2024-07-09
Simplify connection settings

joshk:
- instead of passing in `device_api_host` and `device_api_port`, you can now do all this via `host`
  - you can pass in just a host, or host and port, and it will default to `wss`
- this also fixes an issue where `ws` wasn't supported (I can now use this on localhost easily)
- I've also changed `device_api_sni` to just `sni`
- this is a breaking change

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,14 +3,15 @@ version: 2.1
 latest: &latest
   pattern: "^1.16.*$"
 
-tags: &tags
-  [
+tags:
+  &tags [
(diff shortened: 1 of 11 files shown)


jjcarstens (commented):
Chatted in a call. Josh is going to alter this so there is a warning using the old config values so its not a hard breaking change

jjcarstens (commented):
I think we might also want a warning for each of the `device_api_*` fields we're trying to replace

joshk:
@jjcarstens wanna review again?

jjcarstens (approved):

PR #216, opened 2024-07-13
Support resolving and sending device location information

joshk:
Recently we added support for NervesHub to do some geoip location lookup for connecting devices.

Sadly this was all crushed by the fact that Hub requires the ability to do TLS termination, and thus the `X-FORWARDED-FOR` header can't be added. There is a possible solution, but its also more complicated, and requires changes to Bandit (https://github.com/mtrudel/bandit/issues/379).

This PR is a bit of me admitting defeat with the previous approach, but wanting to make some progress so we can show the device location on a map.

This PR allows the developer to resolve the devices location by implementing `request_location/0` on the `Client` module. I've also included a default which uses the new 'whenwhere.nerves-project.org` service. This information is sent after a successful join, and can switched off within the config.

I've also added support for Hub to ask Link for the current location.


lawik:
I was thinking locations would go in health updates. But probably better with it being a separate mechanism and I see you've included a source field to make it nerves_hub_web's problem to parse it.

If you want a no-deps library for whenwhere I just published one to hex btw. It is called whenwhere.

What I want is to ensure we have an easy way to swap the geo source, so GPS or LTE or whatever. 

If this had a mechanism for defining a source in config with a 0-arity function we call to get the info. Set it to false to explicitly disable. Whenwhere can be default I think?

I think it should send lat/lng so hub doesn't have to know how to deal with various sources.

i started this but based on health checks, so that's why I have the whenwhere lib.

thoughts? :)

joshk:
Great ideas. 

Either we could make the default location resolver be an MFA tuple setting in the config, or we would add this to the client behaviour with a default in the default client. 

And maybe I should simplify this to only lat lng and source.

joshk:
I can also use your lib instead of Req. 

Would or should something like this live in MOTD in the future?

joshk:
@lawik I've updated the PR to isolate the location resolution code in `Client.Default`, as well as specifying the behaviour in the `Client` spec.

I'm getting an odd dialyzer failure, which I don't know how to fix.

I haven't switched to your whenwhere lib yet, I'll do that a little later, this was mostly focused on getting feedback on the approach.

lib/nerves_hub_link/client/default.ex
```diff
@@ -77,4 +77,29 @@ defmodule NervesHubLink.Client.Default do
...
   end
+
+  @impl NervesHubLink.Client
+  def request_location() do
+    case Whenwhere.asks() do
+      {:ok, resp} ->
+        payload = %{
+          source: "whenwhere",
+          latitude: Float.parse(resp.body["latitude"]),
+          longitude: Float.parse(resp.body["longitude"])
```
fhunleth, line 88: It seems like the `Float.parse/1` calls should be in the `Whenwhere` library. `Float.parse/1` returns a tuple.

fhunleth:
I heard that this PR now depends on the extension mechanism. If it does, then I think it's safe to close this. If not and this PR depends on the `whenwhere` library, then I'd like to take a couple passes at that library to make sure that the API is stable. I had a short conversation with @lawik on Slack about returning lat/long as floats rather than strings and not hardcoding the server. There may be other things to add since VintageNet and NervesTime are also potential users that have an alternative implementation at the moment. 

Given that people will be out next week, I'm not sure it's realistic to try to get this in for the 2.5.0 release (#221).

lawik:
Yeah, we pin down down the extension mechanism first.

Then we make a separate geo library.
At least that's what I believe we agreed on :D

lib/nerves_hub_link/client/default.ex
joshk: urg, can't believe i missed that

joshk:
I'll create a new lib for this functionality, using the extension system.

Thanks for the review.

joshk:
I've moved functionality to https://github.com/nervescloud/nerves_hub_geo

PR #220, opened 2024-07-14
Extension pubsub

lawik:


lawik:
Last commit doesn't necessarily make sense. The tuples are various sizes which is not ideal for this. Maybe we should make a map or struct that can be consistently matched on.

Was thinking on the extension-end:

```
  @impl GenServer 
  def handle_info({:broadcast, :msg, topic, event, params}, state) do

    {:noreply, state}
  end
  
  def handle_info({:broadcast, _, _, _}) do
    {:noreply, state}
  end
```

To be able to ignore expected but uninteresting msgs. But that won't work with varied tuples.

lib/nerves_hub_link/pub_sub.ex
```diff
@@ -0,0 +1,87 @@
...
+      # https://hexdocs.pm/elixir/1.17.2/Registry.html#module-using-as-a-pubsub
+      partitions: System.schedulers_online()
+    )
+  end
+
+  @doc """
+  Publish to `NervesHubLink.Socket` to pass up to hub.
+  """
+  def publish_to_hub(topic, event, params) do
+    publish("special:hub", {:to_hub, topic, event, params})
```
joshk, line 25: is there a particular reason for the choice of `"special:hub"`?

lib/nerves_hub_link/pub_sub/message.ex
```diff
@@ -0,0 +1,37 @@
+defmodule NervesHubLink.PubSub.Message do
+  defstruct type: nil, topic: nil, event: nil, params: nil, reply: nil, reason: nil
+
+  alias __MODULE__, as: M
+
+  @type t :: %M{
+          type: :msg | :join | :close | :disconnect,
+          topic: String.t(),
+          event: String.t() | nil,
```
jjcarstens, line 9: The other thing here is I don't think we need to confine ourselves to the Phoenix paradigm of pubsub using strings everywhere that aren't easily checked with dialyzer. Or require vague/generic topics and events. Instead, we could be much more structured with events as atoms or structs themselves and much more explicit.

Subscriptions could be to specific events rather than a string like
```elixir
NervesHubLink.PubSub.subscribe(:my_event)

# or
NervesHubLink.PubSub.subscribe(MyEvent)
```

lib/nerves_hub_link/pub_sub.ex
```diff
@@ -0,0 +1,75 @@
+defmodule NervesHubLink.PubSub do
+  @moduledoc """
+  PubSub mechanism on top of basic Elixir Registry primitives.
+
+  Wrapped up in a module for convenience and convenient correct use in
+  libraries that extend NervesHubLink via this mechanism.
```
jjcarstens, line 6: Do we want to allow external use of this PubSub? I was imagining this as one-way. Link is the only publisher, and everything external are subscribers. For the external extensions to interact with Link, it requires functions to be defined and called directly

lib/nerves_hub_link/socket.ex
```diff
@@ -257,7 +274,14 @@ defmodule NervesHubLink.Socket do
   ##
   # Device API messages
   #
-  def handle_message(@device_topic, "fwup_public_keys", params, socket) do
+  # handle_message captures all messages and does PubSub publishing
+  # hands off handling to `handle_event/3`
+  def handle_message(topic, event, params, socket) do
+    PubSub.publish_channel_event(topic, event, params)
+    handle_event(topic, event, params, socket)
+  end
```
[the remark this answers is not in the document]
lawik: I feel like there were two major points to the extension mechanism:

1. Get extensions out of the NervesHubLink app/supervision tree to ensure connection and firmware updates are not affected by less important communication.
2. Get extensions and lively areas of work out of NervesHubLink to limit how much code review pressure we put on limited maintainer resources.

If we don't use a generic mechanism we lose a decent bit of number 2 as we then end up reviewing which messages should be passed through. Maybe there is some upside to that which I'm not feeling currently?

As a bonus the general mechanism allows implementing custom events based on stuff happening in NervesHubLink, such as easily showing firmware update progress in UI and so on. I don't believe there is an easy way to hook into Link events today and we definitely want to as we go.

My previous mechanism had a system for extensions requesting specific events only but that was definitely more involved.

But are you thinking that we add a `PubSub.public_channel_event/3` to handle_message callbacks explicitly? I did this variant to not screw up anything currently implemented and make the event broadcast truly transparent.

Open to changing it but concerned to lose some of the upside.

lib/nerves_hub_link/pub_sub/message.ex
```diff
@@ -0,0 +1,37 @@
+defmodule NervesHubLink.PubSub.Message do
+  defstruct type: nil, topic: nil, event: nil, params: nil, reply: nil, reason: nil
+
+  alias __MODULE__, as: M
```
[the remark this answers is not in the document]
lawik: We certainly can. I find it a bit verbose and annoying to read personally so that's just a style-thing on my end.

lib/nerves_hub_link/socket.ex
```diff
@@ -459,6 +483,8 @@ defmodule NervesHubLink.Socket do
 
   @impl Slipstream
   def handle_topic_close(topic, reason, socket) when reason != :left do
+    PubSub.publish_topic_close(topic, reason)
```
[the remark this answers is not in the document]
lawik: They should probably know that they won't be receiving more messages. Extensions are likely to be stateful and being aware that they are not open to receive messages is likely relevant I figure.

lib/nerves_hub_link/socket.ex
```diff
@@ -468,6 +494,8 @@ defmodule NervesHubLink.Socket do
 
   @impl Slipstream
   def handle_disconnect(reason, socket) do
+    PubSub.publish_disconnect(@device_topic, reason)
+    PubSub.publish_disconnect(@console_topic, reason)
```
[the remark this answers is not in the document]
lawik: If you want alarms to be sane it requires bringing in something like Alarmist which we may end up doing by default but alarms by default are not good for notification.

lib/nerves_hub_link/pub_sub/message.ex
lawik: If we are only passing through events without looking at them I think following the Phoenix way makes most sense. If we are going to be more explicit in Socket, then making atom events would be better, yeah :)

lib/nerves_hub_link/pub_sub.ex
lawik: We could make publish private since we changed pushing messages up hubwards to be a GenServer.call. It wasn't, earlier.

I want extensions to have access to a subscribe function at least.

jjcarstens:
Sorry this has been so dormant. I chatted with Josh about some semi coherent ideas and landed on a mixture of things here. He recommended I make a commit showcasing what I'm thinking instead so I'm going to work on that in the next couple days

lawik:
Okay, so this would give clear separation between main link and extensions.
It would prevent extensions from reacting to the normal link events/messages which I guess is fine but a bit dull :)

I don't think we are likely to mix in other apps on the server side. Kind of cool but not needed.

Is the primary win the scoping into separate channels and prefixes? I guess we do get some flexibility in routing/targeting on the server side as well. Could come in handy later.

lawik:
Work continues here: https://github.com/nerves-hub/nerves_hub_link/pull/228

PR #225, opened 2024-08-06
Fix problem where signed_at time was often expired

lawik:
When rebooting a device NervesTime will try to restore a reasonable time from the RTC or the FileTime simulated RTC. This then sets OS time.

The shared secret signing code was pulling System.system_time/1 which can lag behind System.os_time/1 for many minutes and can often start on a time that is wildly out of whack.

This changes to use os_time meaning devices should reboot and connect much faster by default.

### DIFF
--- a/lib/nerves_hub_link/configurators/shared_secret.ex
+++ b/lib/nerves_hub_link/configurators/shared_secret.ex
@@ -29,7 +29,9 @@ defmodule NervesHubLink.Configurator.SharedSecret do
       |> Keyword.put_new(:key_length, 32)
       |> Keyword.put_new(:signature_version, "NH1")
       |> Keyword.put_new(:identifier, Nerves.Runtime.serial_number())
-      |> Keyword.put(:signed_at, System.system_time(:second))
+      # Important to use os_time as system_time is not updated by NervesTime
+      # nearly as quickly
+      |> Keyword.put(:signed_at, System.os_time(:second))
 
     alg =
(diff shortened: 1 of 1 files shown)


lib/nerves_hub_link/configurators/shared_secret.ex
```diff
@@ -29,7 +29,9 @@ defmodule NervesHubLink.Configurator.SharedSecret do
       |> Keyword.put_new(:key_length, 32)
       |> Keyword.put_new(:signature_version, "NH1")
       |> Keyword.put_new(:identifier, Nerves.Runtime.serial_number())
-      |> Keyword.put(:signed_at, System.system_time(:second))
+      # Important to use os_time as system_time is not updated by NervesTime
+      # nearly as quickly
```
fhunleth, line 33:
```suggestion
      # Important to use os_time as system_time is not updated by Erlang as
      # quickly. See https://www.erlang.org/doc/apps/erts/time_correction#erlang-system-time
```

fhunleth (approved):
This makes sense. I didn't appreciate how long system time would take to converge with os time, but I guess we know now that it's definitely not instantaneous.

lawik:
Yeah, I've seen it take a decent bit. And it is aggravated by it trying a few times and then backing off. So it takes some time before it will even try again.

lawik:
I don't have merge on this repo btw.

PR #243, opened 2024-11-13
Enable running as a standalone release with runtime configuration

lawik:


lawik:
Is this roughly what you had in mind for running this standalone @jjcarstens? I've been doing it by running a modified dev.exs but I figure we should fix up the way of running it in this way :)

This should help people that want to try NervesHub on not-Nerves such as Joao with the whole Yocto thing.

It should have more config and I need to try more stuff with it but wanted to check that we are thinking roughly the same.

jjcarstens (commented):
`runtime.exs` is the easiest for standalone and allowing using application code before start (like X509 for reading certs), but I don't think one should be committed to the repo. I think `runtime.exs` should be added to the gitignore and we should have examples to copy (either in README or `runtime.exs.example`?). That way users have a path to get the standalone running without having to keep up with custom env vars. What do you think?

lawik:
Some way of passing runtime config I think is the important part and if the standalone is intended to be used in circumstances where people don't really do Elixir? Maybe even precompiled? I feel like runtime.exs is weirdo mystery-meat for most people. So having one which let's you configure things either with ENV vars or some basic config file would be nice I think.

I am okay with gitignore and an example to work from but it's not super compelling either. A good default runtime.exs would also help people wanting to hack on nerves_hub_(web|link) and needs to fake a device.

Just set a direnv type thing for NH_HOST, NH_PRODUCT_KEY and NH_PRODUCT_SECRET.

jjcarstens:
The expected use of NHL is with Elixir and as a dependency. I view the runtime.exs as an experimental tool in this case which enables development and small local experiments. I think past that is too far from scope, especially since we have the Configurator behavior which is intended to use when needing to customize behavior on device. The `runtime.exs` is merely an example of another way you can achieve simple configuration.

lawik:
I know for NervesCloud we want to be able to ship an agent for people that are not ready for a Nerves system yet but could use the firmware delivery and fwup with what they have. We could wrap NervesHubLink in a separate project to do that. You were the one that said you could just run NervesHubLink directly so that's what I went to explore :D 

Maybe you didn't intend an agent at the end. But I'm fairly sure we'll need to do that at some point, happy to put that design in a separate thing that depends on NHL. Figured this approach was more useful overall.

jjcarstens:
I guess I'm still calling that "experimental"

There are 2 confusing things I'm saying here 😜 :
* The `runtime.exs` is one of the easiest tools to use with NHL for quickly connecting, especially dev
* A runtime.exs flavor with lots of custom bits should not be officially maintained

The crux is there will always be very specific bits to add to get connected. Shared secrets, certificate, host, etc. We can make custom env vars, but now that is yet another layer on top of the configuration needed vs a sample file where you fill in the lines and don't have to worry about it being committed

In your agent setup, you may even have more specific changes needed for NervesCloud. You can use this runtime approach mixed with whatever custom bits but in your own manner outside OSS maintenance

lawik:
Fair enough :)

We can own that part.

PR #249, opened 2024-11-22
Report that device is rebooting after successful firmware update

nshoes:
Reason: https://github.com/nerves-hub/nerves_hub_web/issues/494

### DIFF
--- a/lib/nerves_hub_link/client.ex
+++ b/lib/nerves_hub_link/client.ex
@@ -181,6 +181,7 @@ defmodule NervesHubLink.Client do
 
       {:ok, 0, _message} ->
         initiate_reboot()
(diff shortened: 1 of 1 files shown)


lib/nerves_hub_link/client.ex
```diff
@@ -181,6 +181,7 @@ defmodule NervesHubLink.Client do
 
       {:ok, 0, _message} ->
         initiate_reboot()
+        NervesHubLink.send_update_status("rebooting_after_update")
```
jjcarstens, line 184: You may want to swap these. You're going to be racing to deliver this message before the shutdown initiates. I can't remember if this is a graceful reboot or not, but may want all the time you can't get

nshoes:
We're going to handle this web-side.

Issue #250, opened 2024-11-28
Health check fails on alarm report


elinol:
Health reports are failing because `:alarm_handler.get_alarms()` can return a single tuple instead of a list which is expected in the alarm report:
```
 @impl Report
  def alarms do
    for {id, description} <- :alarm_handler.get_alarms(), into: %{} do
      try do
        {inspect(id), inspect(description)}
      catch
        _, _ ->
          {"bad alarm term", ""}
      end
    end
  end
```
This prevents the health report from proceeding with the other reports like metrics and connectivity.

On a test device:
```
iex(39)> NervesHubLink.Extensions.Health.DefaultReport.alarms
** (Protocol.UndefinedError) protocol Enumerable not implemented for {:error, :bad_module} of type Tuple. This protocol is implemented for the following type(s): CircularBuffer, Date.Range, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, Jason.OrderedObject, List, Map, MapSet, Range, Stream
    (elixir 1.17.2) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir 1.17.2) lib/enum.ex:166: Enumerable.reduce/3
    (elixir 1.17.2) lib/enum.ex:4423: Enum.reduce/3
    (nerves_hub_link 2.5.2) lib/nerves_hub_link/extensions/health/default_report.ex:25: NervesHubLink.Extensions.Health.DefaultReport.alarms/0
    iex:39: (file)
iex(39)> :alarm_handler.get_alarms()
{:error, :bad_module}
```

joshk:
Fixed!

PR #251, opened 2024-11-28
Rescue errors in alarms report

elinol:
Fix for issue #250 

This will rescue any exception when creating alarm report for health checks.

It won't solve the problem where `:alarm_handler.get_alarms` returns `{:error, :bad_module}` though.
Any input on that is very welcomed.

### DIFF
--- a/lib/nerves_hub_link/extensions/health/default_report.ex
+++ b/lib/nerves_hub_link/extensions/health/default_report.ex
@@ -30,6 +30,9 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
           {"bad alarm term", ""}
       end
     end
+  rescue
+    error ->
+      %{NervesHubLink.AlarmReportFailed => inspect(error)}
   end
 
   @impl Report
--- a/lib/nerves_hub_link/extensions/health/default_report.ex
+++ b/lib/nerves_hub_link/extensions/health/default_report.ex
@@ -22,17 +22,13 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
 
   @impl Report
   def alarms do
-    for {id, description} <- :alarm_handler.get_alarms(), into: %{} do
-      try do
-        {inspect(id), inspect(description)}
-      catch
-        _, _ ->
-          {"bad alarm term", ""}
-      end
+    case :alarm_handler.get_alarms() do
+      alarms when is_list(alarms) ->
+        for {id, description} <- alarms, into: %{}, do: {inspect(id), inspect(description)}
+
+      err ->
+        %{"NervesHubLink.AlarmReportFailed" => inspect(err)}
     end
-  rescue
-    error ->
-      %{NervesHubLink.AlarmReportFailed => inspect(error)}
   end
 
   @impl Report


lib/nerves_hub_link/extensions/health/default_report.ex
```diff
@@ -30,6 +30,9 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
           {"bad alarm term", ""}
       end
     end
+  rescue
+    error ->
+      %{NervesHubLink.AlarmReportFailed => inspect(error)}
```
jjcarstens, line 35: This breaks the Report spec of `%{String.t() => String.t()}`. But also, I think a simple case here would be simpler and handle situations where a potentially replaced alarm handler does not return expected result

```elixir
case :alarm_handler.get_alarms() do
  alarms when is_list(alarms) ->
    for {id, description} <- alarms, into: %{}, do: {inspect(id), inspect(description)}
  err ->
    %{"NervesHubLink.AlarmReportFailed" => inspect(err)}
end
```

There should not be a need to catch `inspect/1` calls. If that were to error, it would be a core problem in Elixir and we'd probably want to know.

elinol: Changed to this suggestion in the latest commit, and it works fine but the dialyzer is complaining, maybe because it expects `:alarm_handler.get_alarms` to always return a list?

fhunleth: I think there needs to be another way. `:alarm_handler.get_alarms/0` seems like it's only for experimenting with the default alarm handler. The default alarm handler is missing so many features (deduplication being a really important one) that perhaps it's better to let the user know that they should install a proper alarm handler if they want this information.

`{:error, :badmodule}` is a successful return value since it indicates that the default alarm handler has been upgraded to a proper one.

There's no generic way of getting the current set of alarms. The only option I can think of is to let the user specify a MFA or function ref to get them. That would also let the user filter alarms if they don't want them all being reported.

elinol: @fhunleth with "let the user specify", would that be via nerveshub web or with a config setting for the nerves project?

lawik: Oh, this never got merged. That was the intent:
https://github.com/nerves-hub/nerves_hub_link/pull/240

I was assuming that `:alarm_handler.get_alams/0` could be trusted under `alarmist` or other reasonable handlers.

lawik: So I think the goal right now is. Grab the code from that PR that checks for the alarm handler being used. If anything but alarmist we just report empty alarms because we don't know what we'll get.

If alarmist, use a good alarmist function for pulling the alarms. @fhunleth I know you were planning changes there. What should we be using now and should we expect a change near-term?

elinol: What about moving the alarm handler check to the default report, like this: #263 ?

lawik: Yes. That :)

lawik: Doest hat make this PR irrelevant? I think it might.

elinol: Yep

joshk:
@elinol @lawik can this PR be closed?

elinol:
> @elinol @lawik can this PR be closed?

I think so yes

PR #252, opened 2024-12-02
Match device_api_version to mix release version

nshoes:


lib/nerves_hub_link/configurator.ex
```diff
@@ -3,7 +3,7 @@ defmodule NervesHubLink.Configurator do
   alias __MODULE__.Config
   require Logger
 
-  @device_api_version "2.1.0"
+  @device_api_version "2.5.2"
```
jjcarstens, line 6:
```suggestion
  @device_api_version Mix.Project.config()[:version]
```
I think you could change this to be the version configured for the project. Alternatively, you can remove this module attribute and use `to_string(Application.spec(:nerves_hub_link, :vsn))` when building the config at runtime

jjcarstens (approved):

PR #263, opened 2025-01-07
Use Alarmist for current alarms in default health report

elinol:


lib/nerves_hub_link/extensions/health/default_report.ex
```diff
@@ -21,14 +21,23 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
...
+        try do
+          {inspect(id), inspect(description)}
+        catch
+          _, _ ->
+            {"bad alarm term", ""}
+        end
       end
+    rescue
+      _ -> %{}
     end
```
elinol, line 41: I don't really like these nested `try`, have to cover for cases when `Alarmist` is not available. Any other takes on this?

elinol: Maybe something like this is better:
```
  @impl Report
  def alarms() do
    # Currently, only Alarmist is supported as alarm handler.
    # Send empty map if Alarmist isn't loaded.
    alarms =
      try do
        apply(Alarmist, :get_alarms)
      rescue
        _ -> %{}
      end

    for {id, description} <- alarms, into: %{} do
      try do
        {inspect(id), inspect(description)}
      catch
        _, _ ->
          {"bad alarm term", ""}
      end
    end
  end
```

Credo is complaining on using `apply` though.

lawik: I think we can use some compile-time power here:

```
if Code.ensure_loaded?(Alarmist) do
  def alarms() do
    # .. the things
  end
else
  def alarms() do
    %{}
  end
end
```

elinol: Cool, I'll give it a try

lawik (approved):

joshk:
@lawik is this good to merge in?

lawik:
It should be better than the previously broken approach, so yeah :)

PR #266, opened 2025-01-21
Fix bad key formatting for custom health metrics

lawik:


lawik:
Ran into this while testing. I don't want the key to be `"my_key`" or `:my_key` :D

joshk:
@lawik I think this has been addressed during my work with the metric sets.

PR #267, opened 2025-01-24
Add support for tracking total network traffic

joshk:
This sends the total bytes sent and received for each interface, except local.

One thing we should think about is how to track the delta. 

A potential issue I anticipate is a connection crashing, making it look like the device has consumed a lot of data from its first report.

I wonder if we should track the previous total in link and only send the delta. If there is no previous total, we send zeros.

But the downside of this is that if a report isn't received and processed, we will miss some data. Sooooo maybe, we send totals, but we cache the first report and use that for the baseline for future reports.


lawik:
For now I think totals is a great start with warts and all. These also wrap back around to 0 at ~4Gb.

We can make something more nuanced later.

This was on my list, super that you got to it :)

lawik:
Since this is one of the default things we offer we can try to be a bit smart and summarize them similar to Load averages or something on the primary Health display (Device Details view). And we can have an information icon to hover that explains the data and how it might behave a little bit. Give some context.

lawik (approved):

PR #272, opened 2025-02-08
Support a new set of firmware update status messages

joshk:
These new accepted Hub messages hook deeper into the updating lifecycle, as well as fixing issues where new firmware aren't sent to devices when an update is ignored or encounters errors.

This also moves the responsibility for rescheduling the update to Hub, which will resend the update shortly after the date sent by Link.

**The new states sent to hub are:**
- `:updating` (the progress is sent with this message)
- `:finalizing` (the download and applying of the update has finished and the device is about to reboot)
- `:ignored` (the device doesn't want the update. We could also allow a 'reason' to be sent)
- `:reschedule` (including the datetime to block device updates until)
- `:error` (something went wrong, a reason is also sent)

**Two things to discuss/consider:**
- I've changed the `finalizing` push to send a reply from hub, which allows us to make sure hub knows we are about to reboot, and thus track if a device updated by never came back online. I think, for consistency, we make all pushes to `firmware_update_status` return replies of `:ok`, and then its up to Link to decide if it cares about them or not.
- FWUP can send out info and warning messages (eg. https://github.com/nerves-project/nerves_system_rpi0/blob/main/fwup.conf#L292), but they are captured as 'warnings'. We could also send these up to Hub for extra data on how the update is going.

**Some bits to track**
- [ ] Update Hub to accept these messages (I have a changeset ready but we first need to merge in the new orchestrator)
- [ ] Test with some devices

### DIFF
--- a/lib/nerves_hub_link/configurator.ex
+++ b/lib/nerves_hub_link/configurator.ex
@@ -169,6 +169,8 @@ defmodule NervesHubLink.Configurator do
       |> Map.put("fwup_version", fwup_version())
       |> Map.put("device_api_version", @device_api_version)
       |> Map.put("console_version", @console_version)
+      |> Map.put("library", "nerves_hub_link")
+      |> Map.put("library_version", to_string(Application.spec(:nerves_hub_link, :vsn)))
 
     %{base | params: params, socket: socket, ssl: ssl, fwup_devpath: fwup_devpath}
   end
--- a/lib/nerves_hub_link/configurator.ex
+++ b/lib/nerves_hub_link/configurator.ex
@@ -169,8 +169,8 @@ defmodule NervesHubLink.Configurator do
       |> Map.put("fwup_version", fwup_version())
       |> Map.put("device_api_version", @device_api_version)
       |> Map.put("console_version", @console_version)
-      |> Map.put("library", "nerves_hub_link")
-      |> Map.put("library_version", to_string(Application.spec(:nerves_hub_link, :vsn)))
+      |> Map.put("client", "nerves_hub_link")
+      |> Map.put("client_version", to_string(Application.spec(:nerves_hub_link, :vsn)))
 
     %{base | params: params, socket: socket, ssl: ssl, fwup_devpath: fwup_devpath}
   end
--- a/lib/nerves_hub_link/configurator.ex
+++ b/lib/nerves_hub_link/configurator.ex
@@ -27,7 +27,7 @@ defmodule NervesHubLink.Configurator do
 
   require Logger
 
-  @device_api_version "2.2.0"
+  @device_api_version "2.3.0"
   @console_version "2.0.0"
 
   defmodule Config do
--- a/lib/nerves_hub_link/socket.ex
+++ b/lib/nerves_hub_link/socket.ex
@@ -40,14 +40,17 @@ defmodule NervesHubLink.Socket do
     GenServer.cast(__MODULE__, :reconnect)
   end
 
-  @spec send_update_progress(non_neg_integer()) :: :ok
-  def send_update_progress(progress) do
-    GenServer.cast(__MODULE__, {:send_update_progress, progress})
+  @spec send_firmware_update_progress(percent_progress :: non_neg_integer()) :: :ok
+  def send_firmware_update_progress(percent_progress) do
+    GenServer.cast(
+      __MODULE__,
+      {:send_firmware_update_status, :progress, %{percent: percent_progress}}
+    )
   end
 
-  @spec send_update_status(String.t()) :: :ok
-  def send_update_status(status) do
-    GenServer.cast(__MODULE__, {:send_update_status, status})
+  @spec send_firmware_update_status(status :: atom(), payload :: map()) :: :ok
+  def send_firmware_update_status(status, payload \\ %{}) do
+    GenServer.cast(__MODULE__, {:send_firmware_update_status, status, payload})
   end
 
   @spec check_connection(atom()) :: boolean()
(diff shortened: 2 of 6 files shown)


lib/nerves_hub_link/socket.ex
```diff
@@ -270,13 +273,39 @@ defmodule NervesHubLink.Socket do
...
+
+    {:noreply, socket}
+  end
+
+  def handle_cast({:send_firmware_update_status, :finalizing, payload}, socket) do
+    {:ok, push_ref} =
+      push(
+        socket,
+        @device_topic,
+        "firmware_update_status",
```
jjcarstens, line 291: This new event name couples itself to a NervesHub update. May not a bad thing, but just calling out for history

lib/nerves_hub_link/update_manager.ex
```diff
@@ -186,43 +191,42 @@ defmodule NervesHubLink.UpdateManager do
...
-  end
+      {:reschedule, minutes, :minutes} ->
+        until =
+          DateTime.utc_now()
+          |> DateTime.add(minutes, :minute)
 
-  defp maybe_update_firmware(_, _, state), do: state
+        NervesHubLink.send_firmware_update_status(:reschedule, %{until: until})
+        Logger.info("[NervesHubLink] rescheduling firmware update in #{minutes} minutes")
+        {:rescheduled, %{state | status: :idle, previous_status: :rescheduled}}
```
jjcarstens, line 219: I don't understand the addition of `previous_status`. It seems to me that where it is used, `:status` is being set to `:idle`. So now we have a split of states that doesn't seem necessary? It seems saying something like `status: :rescheduled` is a valid state to sit in.

lib/nerves_hub_link/update_manager.ex
```diff
@@ -186,43 +191,42 @@ defmodule NervesHubLink.UpdateManager do
...
 
-  defp maybe_cancel_timer(%{update_reschedule_timer: nil} = state), do: state
-
-  defp maybe_cancel_timer(%{update_reschedule_timer: timer} = state) do
-    _ = Process.cancel_timer(timer)
+      {:reschedule, ms} ->
+        until =
+          DateTime.utc_now()
+          |> DateTime.add(round(ms / 1_000), :second)
```
jjcarstens, line 224:
```suggestion
          |> DateTime.add(ms, :millisecond)
```
I don't think you need to round. Just add the expected time here

lib/nerves_hub_link/configurator.ex
```diff
@@ -156,6 +156,8 @@ defmodule NervesHubLink.Configurator do
       |> Map.put("fwup_version", fwup_version())
       |> Map.put("device_api_version", @device_api_version)
       |> Map.put("console_version", @console_version)
+      |> Map.put("library", "nerves_hub_link")
+      |> Map.put("library_version", to_string(Application.spec(:nerves_hub_link, :vsn)))
```
jjcarstens, line 160:
```suggestion
      |> Map.put("client", "nerves_hub_link")
      |> Map.put("client_version", to_string(Application.spec(:nerves_hub_link, :vsn)))
```
It might be more applicable to call this `client` instead of library?

jjcarstens (commented):
Sending of the messages seems fine to me. I don't have much more opinion on which messages or their names 👍🏼 

I am confused about the `rescheduled` behavior though - It removed all timers and _actual_ rescheduling. Is this intentional? It seems we would still want to have a step we try again or ask NervesHub for a new URL? We might even be intelligent to know valid time length of the URL and if we reschedule within range, we just retry on device without network?

joshk:
Thanks for the feedback @jjcarstens, I'll get onto the code bits tomorrow.

Regarding:

> I am confused about the rescheduled behavior though - It removed all timers and actual rescheduling. Is this intentional? It seems we would still want to have a step we try again or ask NervesHub for a new URL? We might even be intelligent to know valid time length of the URL and if we reschedule within range, we just retry on device without network?

Moving rescheduling to Hub gives us 3 main benefits:
1. The device gets a fresh firmware download URL. This is very, very useful if the reschedule is close to an hour away, ensuring devices aren't using a download URL that will expire and leave the download to fail.
2. Rescheduling releases the concurrent update slot (`InflightUpdate`) being taken by the device. The current implementation blocks other devices from using this slot until the device updates or the `InflightUpdate` is cleared by Hub based on the deployment clean up code.
3. Deployments could be canceled or paused or rolled back, and this change allows Hub to abandon an update to a device that has asked for it to be rescheduled.

Plus we get better metrics and data around the number of devices that have asked for a delay in processing an update.

lib/nerves_hub_link/configurator.ex
joshk: Love this.

lib/nerves_hub_link/update_manager.ex
joshk: True true. For some reason I didn't click that I could use `:millisecond`

lib/nerves_hub_link/socket.ex
joshk: Sorry, can you go a bit deeper into this, I'm not sure what the downside is you are referring to.

lib/nerves_hub_link/update_manager.ex
joshk: My intention was to be able to see what the current status of the `UpdateManager` was, along with the result from the previous request.
eg. is it doing something or is it available to process an update, and what was the outcome of the last request.

I could remove `previous_status` and just use the new simplified statuses of `idle`, `{:updating, integer()}`, and `:applying`, and rely on message sending for the result of an update request 
(eg. `:ignored`, `:rescheduled`, `:successful`, and `{:error, String.t()}`)

PR #279, opened 2025-03-14
Health enhancements

joshk:
Introducing the concept of `MetricSet`s.

Metrics are now split into different sets of data which allows device code to specify the metric data they want to send to NervesHub, while also providing a clear and simple way to define custom metrics, or use metric sets from other libs.

I also improved how string keys are normalized. Previously they were being `inspect`ed, which caused extra escaping where it wasn't needed. eg. if the key was "hello", `inspect` would return `"\"hello\""` , which caused `"hello"` to be the key in the metrics UI.

This PR is thanks to @gworkman after seeing how he integrated his boards battery metrics into his Link config.

The `README.md` updates have a good summary of the change, but for the sake of the clear PR, the concept allows for:

```elixir 
config :nerves_hub_link,
  health: [
    metric_sets: [
      :defaults, # expanded within the NervesHubLink.Extensions.Health code
      MyApp.HealthMetrics, # custom app metrics
      ALibrary.BatteryMetrics # metrics from a lib
  ]
```

### DIFF
--- a/lib/nerves_hub_link/extensions/health/default_report.ex
+++ b/lib/nerves_hub_link/extensions/health/default_report.ex
@@ -62,18 +62,22 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
   end
 
   defp get_metric_sets() do
-    Application.get_env(:nerves_hub_link, :health, [])
-    |> Keyword.get(:metric_sets, @default_metric_sets)
-    |> Enum.map(fn metric_set ->
-      if metric_set in [Default, Defaults] do
-        @default_metric_sets
-      else
-        metric_set
-      end
-    end)
-    |> List.flatten()
-    |> Enum.map(& &1.metrics())
-    |> Enum.reduce(%{}, &Map.merge/2)
+    metric_sets =
+      Application.get_env(:nerves_hub_link, :health, [])
+      |> Keyword.get(:metric_sets, @default_metric_sets)
+      |> Enum.map(fn metric_set ->
+        if metric_set in [:default, :defaults] do
+          @default_metric_sets
+        else
+          metric_set
+        end
+      end)
+      |> List.flatten()
+
+    for metric_set <- metric_sets,
+        {k, v} <- metric_set.sample(),
+        into: %{},
+        do: {normalize_key(k), v}
   end
 
   defp vof({mod, fun, args}), do: apply(mod, fun, args)
@@ -88,7 +92,7 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
     metadata = get_health_config(:metadata, %{})
 
     for {key, val_or_fun} <- metadata, into: %{} do
-      {inspect(key), vof(val_or_fun)}
+      {normalize_key(key), vof(val_or_fun)}
     end
   end
 
--- a/lib/nerves_hub_link/extensions/health/default_report.ex
+++ b/lib/nerves_hub_link/extensions/health/default_report.ex
@@ -78,7 +78,7 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
     |> format_metrics_or_metadata()
   end
 
-  def format_metrics_or_metadata(metrics_or_metadata) do
+  defp format_metrics_or_metadata(metrics_or_metadata) do
     metrics_or_metadata
     |> List.wrap()
     |> normalize_metrics_from_config()
--- a/lib/nerves_hub_link/extensions/health/default_report.ex
+++ b/lib/nerves_hub_link/extensions/health/default_report.ex
@@ -4,8 +4,14 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
   new metadata, metrics and such via config.
   """
   @behaviour NervesHubLink.Extensions.Health.Report
+
   alias NervesHubLink.Extensions.Health.Report
-  require Logger
+
+  @default_metric_sets [
+    NervesHubLink.Extensions.Health.MetricSet.CPU,
+    NervesHubLink.Extensions.Health.MetricSet.Disk,
+    NervesHubLink.Extensions.Health.MetricSet.Memory
+  ]
 
   @impl Report
   def timestamp() do
@@ -36,11 +42,7 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
   def metrics() do
     [
       metrics_from_config(),
-      cpu_temperature(),
-      cpu_utilization(),
-      load_averages(),
-      memory(),
-      disk()
+      get_metric_sets()
     ]
     |> Enum.reduce(%{}, &Map.merge/2)
   end
@@ -59,8 +61,22 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
     |> Enum.reduce(%{}, &Map.merge/2)
   end
 
+  defp get_metric_sets() do
+    Application.get_env(:nerves_hub_link, :health, [])
+    |> Keyword.get(:metric_sets, @default_metric_sets)
+    |> Enum.map(fn metric_set ->
+      if metric_set in [Default, Defaults] do
+        @default_metric_sets
+      else
+        metric_set
+      end
+    end)
+    |> List.flatten()
+    |> Enum.map(& &1.metrics())
+    |> Enum.reduce(%{}, &Map.merge/2)
+  end
+
   defp vof({mod, fun, args}), do: apply(mod, fun, args)
-  defp vof(fun) when is_function(fun), do: fun.()
   defp vof(val), do: val
 
   defp get_health_config(key, default) do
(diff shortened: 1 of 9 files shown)


lawik (approved):

lawik:
Sounds quite close to what already existed but good stuff, particularly the stringification changes.

gworkman:
Looks good - thanks for getting this in!! I'll be testing shortly. My only question is about the fail-mode behavior. As in, when there is no internet connection it looks like this send_report will happily report success?

It seems like the spec says push/2 returns an :ok/:error tuple, so maybe we can check if it sent

```elixir
@impl NervesHubLink.Extensions
  def handle_event("check", _msg, state) do
    case push("report", %{"value" => check_health()}) do
        {:ok, _} -> {:noreply, %{state | report_sent: true}}
        {:error, _reason} -> {:noreply, %{state | report_sent: false}}
    end
  end
```

joshk:
@lawik yep, very close. The one big enhancement is that libs, like https://hex.pm/packages/soleil, can provide a single MFA/function that fetches and formats all the possible health stats, which simplifies the inclusion of the stats in the config.

joshk:
> Looks good - thanks for getting this in!! I'll be testing shortly. My only question is about the fail-mode behavior. As in, when there is no internet connection it looks like this send_report will happily report success?
> 
> It seems like the spec says push/2 returns an :ok/:error tuple, so maybe we can check if it sent
> 
> ```elixir
> @impl NervesHubLink.Extensions
>   def handle_event("check", _msg, state) do
>     case push("report", %{"value" => check_health()}) do
>         {:ok, _} -> {:noreply, %{state | report_sent: true}}
>         {:error, _reason} -> {:noreply, %{state | report_sent: false}}
>     end
>   end
> ```

Perfect! I'll add this now!

Are you then able to test this branch with your project?

gworkman:
@joshk yes I'll test in the morning here. Will report back

jjcarstens (commented):
Some general code pieces for a little optimization since there is so much iterating.

I am behind on the state of Health reporting though and this is showing other things that are a little more concerning. The original thoughts of Extensions was to be more controlling and defined in the structure of device <-> server communication, especially with extra bits. But it seems Health is taking the opposite approach and becoming more and more generic. So much that now it seems we are defining behavior and functional code in configuration, all the way down to the level and depth of messages sent to server. 

That aside and continuing the more general approach, there is already a mechanism for defining your own reporting with the `:report` key. If the user wants to add more data (even though probably not handled by NH), then I would expect they make `MyHealthReport` and potentially re-use `DefaultReport` functions to fill in the data within their module.

All that aside, we are this far and I don't think it should block anything. But food-for-thought going forward thinking about how to build this out more. In fact, the counter to my argument is that if this is merely a means of getting fluid data to the server so that it can offload to something better suited to handling metrics (like a data dog or statsd or telemetry sidecar etc), then we would want it to be fairly open. Though I would still suggest not trying to handle everything in configuration and `DefaultReporter` 😉

gworkman:
Tested the metrics changes this morning - I really like the `metrics_set` interface. Being able to define it in a module with callback feels like the right structure to me.

`NervesHubLink.Extensions.Health.send_report/0` works as expected, though the UI doesn't auto-refresh the data (not a big deal though). I'll be testing `NervesHubLink.Extensions.Health.report_sent?/0` shortly, to see if I can get my devices to wait to sync with NervesHub before sleeping

String formatting also looks much better - see screenshot below.

My repo for testing is [here](https://github.com/protolux-electronics/soleil_demo/commit/5a02dd6543c82f9a6eecac85331b63ae95c1f4b3)

Screenshot 2025-03-16 at 12 40 28 PM

gworkman:
I did some testing on the `send_report/0` and `report_sent?/0` functions. I'm running into a few edge-case issues here, since I'm operating just as the entire application is coming up. First, it seems like the GenServer for `NervesHubLink.Extensions.Health` (which is registered and referenced in `health.ex` via `name: __MODULE__`) isn't started until after the socket is connected and configuration is received from the server:

```
12:30:16.903 [info] [NervesHubLink] waiting for network to become available

12:30:17.171 [info] RouteManager: set_connection_status wlan0 -> :internet (timeout)

12:30:18.904 [info] [NervesHubLink] waiting for network to become available

12:30:20.914 [info] [NervesHubLink] connecting to devices.nervescloud.com

12:30:21.434 [info] [NervesHubLink] connection to devices.nervescloud.com succeeded

12:30:21.435 [info]     :alarm_handler: {:clear, NervesHubLink.Disconnected}

12:30:21.541 [error] Task #PID<0.1792.0> started from SoleilDemo.Supervisor terminating
** (stop) exited in: GenServer.call(NervesHubLink.Extensions.Health, :report_sent?, 5000)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
    (elixir 1.18.2) lib/gen_server.ex:1121: GenServer.call/3
    (soleil_demo 0.1.0) lib/soleil_demo/battery_logger.ex:58: SoleilDemo.BatteryLogger.send_nerveshub_report/2
    (soleil_demo 0.1.0) lib/soleil_demo/battery_logger.ex:20: SoleilDemo.BatteryLogger.run/1
    (elixir 1.18.2) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2
Function: &SoleilDemo.BatteryLogger.run/1
    Args: [[]]
```

So maybe a quick check if the process is alive first would be in order. And I also think the `send_report/0` function would be better as a `call` rather than `cast` - thoughts on this? It would let me avoid the mess below:

```elixir
  defp send_nerveshub_report(tries \\ 100, timeout \\ 100)

  defp send_nerveshub_report(0, _timeout), do: {:error, :timeout}

  defp send_nerveshub_report(tries, timeout) do
    if NervesHubLink.socket_connected?() do
      NervesHubLink.Extensions.Health.send_report()

      # wait for 50 ms for the report to send
      Process.sleep(50)

      if NervesHubLink.Extensions.Health.report_sent?() do
        :ok
      else
        send_nerveshub_report(tries - 1, timeout)
      end
    else
      Process.sleep(timeout)
      send_nerveshub_report(tries - 1, timeout)
    end
  end
```

joshk:
Thanks for the feedback @gworkman 

The GenServer for Health is only started after Hub tells the Device to connect its extensions.

To aid in knowing when the channel has connected, I've added `NervesHubLink.Socket.extensions_connected?()` which you can poll.

I've also switched the `send_report/0` from `cast` to `call`.

gworkman:
Just tested - works like a charm! :smile:

[Here's the commit](https://github.com/protolux-electronics/soleil_demo/commit/d8243a4bcf1336785c6bcc9fa0b95acd81009aea) if you want to see the implementation

joshk:
@gworkman awesome!

It would be cool to have https://github.com/protolux-electronics/soleil_demo/commit/5a02dd6543c82f9a6eecac85331b63ae95c1f4b3#diff-77d77e7500bcf972c54a25d1a3490f07ed21cf8be11022feb78def06e856a95d in the `soleil` lib, then adding the metrics would be as simple as:

```
    metric_sets: [Defaults, Soleil.Metrics]
```

gworkman:
@joshk - that's the plan! I am working on moving it over :)

joshk:
> @joshk - that's the plan! I am working on moving it over :)

Faaaaaantastic! Super super cool!

lib/nerves_hub_link/extensions/health/default_report.ex
```diff
@@ -59,6 +61,21 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
     |> Enum.reduce(%{}, &Map.merge/2)
   end
 
+  defp get_metric_sets() do
+    Application.get_env(:nerves_hub_link, :health, [])
+    |> Keyword.get(:metric_sets, @default_metric_sets)
+    |> Enum.map(fn metric_set ->
+      if metric_set in [Default, Defaults] do
```
jjcarstens, line 68: Not sure how I feel about this. `Default` and `Defaults` aren't real modules and is effectively the same as `:defaults`, except maybe more misleading since someone might go searching for a non-existent module.

I would vote for not using it and require being explicit. Have the defaults be used if no `:metric_sets` seems appropriate as one could then explicitly set it to `nil` or `[]` to skip those metrics. If we _need_ to have an easy defaults key, then I think I need to relent and go with `:defaults` to prevent someone needlessly digging for a non-existent module

lib/nerves_hub_link/extensions/health/metric_set.ex
```diff
@@ -0,0 +1,7 @@
+defmodule NervesHubLink.Extensions.Health.MetricSet do
+  @moduledoc """
+  Behaviour for implementing a custom metric set to be used in a report.
+  """
+
+  @callback metrics() :: %{String.t() => number()}
```
jjcarstens, line 6:
```suggestion
  @callback metrics() :: %{String.t() | atom() => number()}
```
This says `String.t()` keys but several of the default metric_set implementations are atom keys. Allowing both would probably work for simplicity, but merging later on needs to be explicit

lib/nerves_hub_link/extensions/health/default_report.ex
```diff
@@ -59,6 +61,21 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
...
+        @default_metric_sets
+      else
+        metric_set
+      end
+    end)
+    |> List.flatten()
+    |> Enum.map(& &1.metrics())
+    |> Enum.reduce(%{}, &Map.merge/2)
```
jjcarstens, line 76: This value from application config + `Enum.reduce` + merging is what is being missed by dialyzer. Here we should probably be more strict in the keys merged. Mixture of atoms and strings will be duplicated over the wire, but deduped on decode on the server leading to wasted bytes

image

One approach would be to filter and transform as you are iterating through the metric sets instead
```suggestion
    for metric_set <- metric_sets, {k, v} <- metric_set.metrics(), into: %{}, do: {to_string(k), v}
```

image

joshk: Great find. I don't love the `for` comprehension, but I've gone with it.

lib/nerves_hub_link/extensions/health/default_report.ex
joshk: I've switched this to `:defaults` and `:default`. I think this is useful to have.

joshk:
@gworkman sorry, slight change to the api, I switched from `metrics/0` to `sample/0`

lib/nerves_hub_link/extensions/health/default_report.ex
```diff
@@ -59,6 +61,25 @@ defmodule NervesHubLink.Extensions.Health.DefaultReport do
...
+        else
+          metric_set
+        end
+      end)
+      |> List.flatten()
+
+    for metric_set <- metric_sets,
+        {k, v} <- metric_set.sample(),
+        into: %{},
+        do: {normalize_key(k), v}
```
jjcarstens, line 80: normalize_key only checks that it is printable and then inspects it. Does that mean atoms would just be converted to `":some_atom"` via inspect? If so, that would leave the same problem of duplicated data with key variations

joshk: There is a version of `normalize_key` that checks if it is not binary and then will use `to_string`

https://github.com/nerves-hub/nerves_hub_link/pull/279/files#diff-b1751145aa12511d57bcf440036ff0cfea3945870a4048019fecd12fd91d0e78R123-R126

jjcarstens (approved):

PR #280, opened 2025-03-16
Fix the config key used for fetching the `retry_config`

joshk:
**And** improve the docs for the `Downloader` and `RetryConfig` modules.

The retry logic was merged in from a lib called `nerves_hub_link_common` many moons ago, but the key used for fetching the config and the version from the spec were not updated.

This fixes the `User-Agent` header, and aligns how the retry logic should be configured.

I've also given the docs for the `RetryConfig` module a new lick of paint, opting to use [`NimbleOptions`](https://hexdocs.pm/nimble_options).

### DIFF
--- a/lib/nerves_hub_link/downloader/retry_config.ex
+++ b/lib/nerves_hub_link/downloader/retry_config.ex
@@ -1,69 +1,77 @@
+# credo:disable-for-this-file Credo.Check.Readability.StrictModuleLayout
 defmodule NervesHubLink.Downloader.RetryConfig do
-  @moduledoc """
-  Configuration structure for how the Downloader server will
-  handle disconnects, errors, timeouts etc
-  """
-
-  defstruct [
-    # stop trying after this many disconnects
-    max_disconnects: 10,
-
-    # attempt a retry after this time
-    # if no data comes in after this amount of time, disconnect and retry
-    idle_timeout: 60_000,
-
-    # if the total time since this server has started reaches this time,
-    # stop trying, give up, disconnect, etc
-    # started right when the gen_server starts
-    # default is 24 hours as that is how long NervesHub AWS urls are signed for
-    max_timeout: 86_400_000,
-
-    # don't bother retrying until this time has passed
-    time_between_retries: 15_000,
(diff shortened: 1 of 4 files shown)


nshoes (commented):
Why bring in another library to validate schemas when we have Ecto?

joshk:
> Why bring in another library to validate schemas when we have Ecto?

I've never seen Ecto uses for type specs. Plus, NimbleOptions was already in the included deps, and it was built for situations like this. Eg. it is used by Slipstream, and was built by Dashbit.

lib/nerves_hub_link/downloader/retry_config.ex
```diff
@@ -1,69 +1,77 @@
...
+          time_between_retries: non_neg_integer(),
+          worst_case_download_speed: non_neg_integer()
         }
+
+  @doc """
+  Validates a proposed configuration, raising on error
+  """
+  @spec validate!(Keyword.t()) :: t()
+  def validate!(opts) do
+    validated = NimbleOptions.validate!(opts, @definition)
```
jjcarstens, line 74: Is there a way to validate and use defaults on invalid? We've lasted this long, but it bothers me that none of these options are evaluated until runtime when trying to download a firmware update. Kind of a circular problem to have a bug downloading an update which you would need to download an update to fix 😬

jjcarstens (commented):
Ecto is really nice, but is a little heavy handed when we already have NimbleOptions which can effectively do the same thing. Putting in the `mix.exs` just shows we're explicitly using it now and I'm okay with that

lib/nerves_hub_link/downloader/retry_config.ex
joshk: Good catch.

The simple fix for the time being is to use `NimbleOptions.validate/2` and pass in the defaults when there is an `{:error, _}`.

Switching to compile time should be in a different PR, as I don't love that we have multiple modules using `Application.get_env`, so I'd like to take a bigger look at this problem.

jjcarstens (approved):

PR #284, opened 2025-03-18
Remove outdated example app

fhunleth:
I was going to update the example app, but I don't think it's been used in a long long time. I didn't see it referenced anywhere and I was thinking that you all (@joshk and @lawik) would have opinions on people's first try at using NervesHub.


joshk (approved):
Agree. If we want to add one back then we can make sure its up to date.

Issue #292, opened 2025-03-27
Alternative to `:public_key.cacerts_get/0` for older OTP versions


nshoes:
https://github.com/nerves-hub/nerves_hub_link/pull/290 has unearthed another `mix compile failure` for OTP 24:

```
warning: :public_key.cacerts_get/0 is undefined or private
  lib/nerves_hub_link/certificate.ex:67: NervesHubLink.Certificate.ca_certs/0
```

This was introduced in https://github.com/nerves-hub/nerves_hub_link/pull/166. I don't know what we should do here. The cert-specific functions for `:public_key` were introduced in OTP 25; there isn't an obvious function to which we can fall back. Maybe we should noop on older OTP versions? Looking for input.

cc: @fhunleth @joshk

joshk:
how about we emit a warning and recommend that they include the CAStore library in their project?

nshoes:
To clarify: emit a warning on < OTP 25 instead of calling that function? A compile-time check could do that.

joshk:
Yep, you could create a new function, eg. `public_certs()` which, when on OTP < 25, the function is just a warning and returns an empty list.

fhunleth:
`whenwhere` isn't my library. I haven't looked at why it requires Elixir 1.15.

joshk:
@lawik ping

lawik:
I don't think whenwhere actually cares about older Elixirs... Will check in on this.

lawik:
Okay, dialing it back to 1.12 as that hits a nimble_parsec dependency. Trying older OTPs < 24 now.

lawik:
Okay, OTP 23 and less is a nuisance to build due to OpenSSL changes. Let me know if you need them :)

lawik:
whenwhere 0.1.2 is published and should address Elixir 1.12 and OTP 24 and the :logger thing.

lawik:
PR for link: https://github.com/nerves-hub/nerves_hub_link/pull/293

nshoes:
Thanks @lawik! Huge help.

PR #294, opened 2025-03-28
Emit a warning when trying to retrieve public certs through Erlang on < OTP 25

nshoes:
This lets us compile with no warnings on OTP 24. Solves #292.

Credo complains about using `apply/3` when knowing the number of arguments, but I think it's usage is appropriate here.

### DIFF
--- a/lib/nerves_hub_link/certificate.ex
+++ b/lib/nerves_hub_link/certificate.ex
@@ -64,7 +64,21 @@ defmodule NervesHubLink.Certificate do
           "[NervesHubLink] Using default system certificates. Requests may fail if the NervesHub server certificate is not signed by a globally trusted CA, or the installed system certificates are old."
         )
 
-        :public_key.cacerts_get()
+        maybe_public_certs()
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


lib/nerves_hub_link/certificate.ex
```diff
@@ -64,7 +64,22 @@ defmodule NervesHubLink.Certificate do
...
         )
 
-        :public_key.cacerts_get()
+        maybe_public_certs()
+    end
+  end
+
+  # TODO: We can drop this check once OTP 25
+  # is our lowest supported version
+  if String.to_integer(System.otp_release()) < 25 do
```
fhunleth, line 73:
```suggestion
  if String.to_integer(System.otp_release()) >= 25 do
```
The comparison looks flipped. Once this is fixed, you should also be able to turn the `apply` to a direct call.

nshoes: Oh no 😭 Thank you, fixed!

fhunleth (approved):
LGTM!

PR #295, opened 2025-03-28
v2.7.2 release

nshoes:


joshk (approved):
👍

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+## [2.7.2] - 2025-03-28
+
+* Fixed
+  * Retrieve disk info differently based on OTP version
+  * Emit a warning when trying to retrieve public certs through Erlang on < OTP 25
+  * Always run 'mix compile --warnings-as-errors' in CI
+
+* Updated
+  * whenwhere 0.1.1 => 0.1.2
+  * nerves_runtime 0.13.7 => 0.13.8
```
fhunleth, line 12: I don't include changes to `mix.lock` files since they aren't shipped in the hex package, so the `nerves_runtime` version bump doesn't need to be included. The `whenwhere` bump is interesting since the `mix.exs` was updated to force the version bump. That means that anyone updating NHL will get the version bump whether they like it or not. Perhaps reword this to "Force whenwhere 0.1.2 to fix compatibility with OTP 24 and earlier".

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## [2.7.2] - 2025-03-28
+
+* Fixed
+  * Retrieve disk info differently based on OTP version
+  * Emit a warning when trying to retrieve public certs through Erlang on < OTP 25
+  * Always run 'mix compile --warnings-as-errors' in CI
```
fhunleth, line 8: I have a general view that the CHANGELOG.md is meant for users of a library and should be written with them in mind. That's why I try to list CHANGELOG.md in the hex package files list so that if you do a hex diff on hex.pm, you get a bullet list of changes first. Library maintainers have the git commit history available to them, so I feel like they're well served by default.  

So far this line about `--warnings-as-errors` on CI, I'd generally not include it since it doesn't affect end users. I think their benefit is in the first two bullets where the issues that caused the breakage were fixed. Make sense?

```suggestion
```

nshoes: Makes sense, fixed! I like it.

CHANGELOG.md
nshoes: Makes sense! Fixed.

fhunleth (approved):

PR #297, opened 2025-04-07
Fix pattern match for :disksup.get_disk_data()

CJRChang:
[:disksup.get_disk_data()](https://www.erlang.org/doc/apps/os_mon/disksup.html#get_disk_data/0) returns {key, total, capacity_percentage} so the current pattern match is broken.


joshk (approved):
Looks great to me. Thanks for the contribution!

@lawik just checking, when you added this did it ever work for you?

CJRChang:
I tracked this one to #291 . [:disksup.get_disk_info()](https://www.erlang.org/doc/apps/os_mon/disksup.html#get_disk_info/0) returns a 4-tuple, which is probably how this came about.

lawik:
I think that takes me off the hook :D 
I do believe I had it working.

joshk:
@CJRChang you rock!
@lawik and you do to!

PR #298, opened 2025-04-07
Simple persisted downloader

lawik:
Using Req this module doesn't spend quite so many lines of code wrangling headers or doing really, really particular fiddling with connection state.

There are definitely things to tighten up and button down but I figured I'd throw it out to get feedback before I put more into it. I'd like to use Req.Test to make tidier tests and some that act more like S3.

I also want to do the plumbing for requesting S3 reauth while also reporting progress. The threads are there. As are stubs for reporting download progress without fwup streaming involved.

Something this does differently from the default downloader is that it essentially resets the retry limit for downloading whenever it makes progress on the download. This should increase the chance of succeeding on a flaky connection is the theory and would probably keep some innocent boxes out of the penalty box.

This is just the downloader. It won't apply the downloaded firmware file. It won't be a ton of code to do that.

I would have used Req's entire retry mechanism but unfortunately that is not range-aware. It can probably be addressed upstream at some point but that's why.


lawik:
Have tested that it works against NervesCloud. Need to update some tests to match the new flow in UpdateManager.

It will currently default to the new one, might not want to keep that. But if it fails to create a file on disk it will fall back to the classic method. Gives a recovery path if a filesystem is acting up.

Ran some tests where I made the download fail randomly every 100-or-so chunks and it would eventually complete successfully which I thought was nice.

lawik:
Sure!
Go for it. Let me know if anything seems weird.

nshoes:
@lawik I am stumped. I'm passing in a unique port number across the whole suite yet `Cowboy` still blows up. I'm unsure how to continue debugging.

lawik:
The tests seem to have a problem where the fact that we run parallel sets of web servers cause some kind of issue. It is weird, inconsistent and hard to pin down. Turning async off for those test modules seems to make them 100% consistent. The problem didn't seem to exist when it was only one module. So some kind of interaction between them most likely. The log would contain errors indicating port binding problems. But we already bind the ports early, so it seems quite weird.

nshoes:
`async: false` crossed my mind but I really wanted to fix it 😭 Finishing my review is next on my plate; it'll be mostly questions 😬

jjcarstens:
I have 2 maybe helpful thoughts:
* The port logic seems complicated. If you use port `0`, then the _system_ will pick an available port and attempt to prevent collisions
* Maybe of the tests start a web server in a describe block. But all tests in the same file run in serial, so it would be more advantageous to start the web server once for each file (maybe in a `setup_all`). Then you are only having to deal with port uniqueness across serial runs, not every single test, so less possibility for collision

jjcarstens:
One more thing: I think `Bandit` is generally easier to work with. using something like this would also be useful. A single Plug could even route based on path or args to what was needed and only start one test web server instead of per process

```elixir
{:ok, pid} = Bandit.start_link(plug: SomePlug, port: 0)
{:ok, {_ip, port}} = ThousandIsland.listener_info(pid)
url = "http://localhost:#{port}"
```

fhunleth:
I skimmed over this. It's feeling like there are a lot of cooks on this issue right now, so I'm feeling like the best thing for me to do is hold off on code comments until the dust settles. 

I don't know how this affects people's plans, but my main comment was about ensuring test coverage across the scenarios we see in the field. I was thinking that a test run that took 10s of minutes would even be fine for coverage even though we'd obviously make that opt-in for our own sanity. Doing this all in Elixir seems fine, but a scriptable proxy like `toxiproxy` seemed useful too. My main worries are that we accidentally regress in the future due to how hard this code is to test and that trying to avoid TCP timeouts to make the tests run quickly blinds us to edge cases. I haven't thought through all of the field scenarios that I think we've run into, but certainly can put some time into it. Perhaps we need to assemble that list before actually writing any tests beyond the new ones that have been added in this PR.

nshoes:
Okay, I got the tests using Bandit. A few tests are still failing, but not because of re-used ports. They are currently skipped because I wanted to see a green run on CI. I know why they're failing but it took awhile to figure out. Solving that next.

lawik:
@jjcarstens I ran across the port 0 suggestion but the people discussing it were unsure how to get the resulting port. That looks good.

I did try some with Bandit but didn't win anything from it right away so reverted.

@fhunleth Happy to code up a more thorough test suite for whatever things y'all have experienced.

I think there is one part of the retry config we don't do in this new variant. It used to check for when the download rate was lower than a particular transfer speed and essentially kill the download/retry based on that. Considering the goal is to make continuous progress I'm not sure that's a good move anymore. We can resume better than the previous version could. Anyone remember the type of issue that was added to cover?

fhunleth:
Yeah, the download rate too low was a bandaid to not kill console access to hubs with painfully slow connections. There's got to be a better way, though. I agree that it needs to be revisited.

lawik:
Oh, interesting! That makes sense.

lawik:
There was some discussion out-of-band around how to test this for a lot of network scenarios. Essentially try to avoid it causing regressions.

Potentially that'd be made easier with the new qemu system.

But we should also just put it in use and make sure it does what it ought to.

joshk:
This PR was an awesome inspiration for #327, which approaches this download persistence problem in a slightly different way.

Thank you Lars for your investigations on this <3 <3 <3

PR #302, opened 2025-04-23
Fix port numbering in tests

lawik:
The machine I run on happens to have something running on port 80. That failed one test.

Then we were assigning essentially random ports in the 6000+ range which I sometimes have a bunch of.

This commit changes the way we grab ports into a recursive one. It will first grab a likely random port but if it is in use it will recurse to find the next open one. It seems to work well.


joshk (approved):
Love it!

PR #303, opened 2025-05-04
Add support for devices to send log messages to NervesHub

joshk:
This new extension enables devices to send log output to NervesHub for processing and storage.

This is **untested** on a device.

Hub PR companion: https://github.com/nerves-hub/nerves_hub_web/pull/2080


lawik:
I really like the feature :)

All the extensions so far are designed to essentially only report on request from the server, to avoid uncontrollable transmission of data, spamming, saturating connections, tripping rate limits and what not.

Logging can produce a ton of messages, especially if something is wrong, but even when it isn't if the device is just an enthusiastic logger. This makes me a bit hesitant to set it to be on by default.

I could see the NervesHub requesting logs on an interval and you essentially get them as batches by grabbing whatever is in RingLogger at the time and sending that as a batch (which would probably be nicer on the Postgres side as well, batch inserts are fast).

I could also see a mode where you visit the logging tab and that activates the logging. Pulling the most recent backlog from RingLogger on the device and then streaming while you watch it. This wouldn't produce a great history unless we also fetch at some interval.

The safety design of extensions kind of hinge on them not sending information without NervesHub requesting it. Which is counter-intuitive from a performance perspective but is reassuring from a fleet management one. While we can disable an extension and NervesHubLink should detach it.

I could see a setting that you can switch on for your product that does stream all the logs but I wouldn't necessarily use it as a default. Especially as logging as default can suddenly run up an LTE bill because you updated nerves_hub_link and defaults changed.

Happy to twist and turn on this because the feature will be a lot nicer than needing to console in to get some logs.

joshk:
Thanks for the great feedback and super relevant points!

I've removed the extension from the defaults, and marked it as `Early Release`. This will allow us to test it with a select number of devices.

I think getting the Hub PR merged in first is most important, as this PR can stay as a branch while we getting everything working well.

lib/nerves_hub_link/extensions/logging/logger_handler.ex
```diff
@@ -0,0 +1,23 @@
+# SPDX-FileCopyrightText: 2025 Josh Kalderimis
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+defmodule NervesHubLink.Extensions.Logging.LoggerHandler do
+  @moduledoc """
+  Send logs to NervesHub.
+  """
+
```
jjcarstens, line 9:
```suggestion
@behaviour :logger_handler
```
I think we should include the behavior here.

lib/nerves_hub_link/extensions/logging/logger_handler.ex
jjcarstens, line 1: I think this file could be combined with `NervesHubLink.Extensions.Logging` - Not critical, but on devices more modules == slower boot up time. Worth keeping in mind at times 😉

lib/nerves_hub_link/extensions/logging/logging.ex
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  @impl NervesHubLink.Extensions
+  def handle_event(_, _msg, state) do
+    {:noreply, state}
+  end
+
+  @impl GenServer
+  def handle_cast({:send_logs, log_payload}, state) do
+    _ = send_logs(log_payload, state)
```
jjcarstens, line 50: Since we are completely ignoring the result and not using the state, I don't think this needs to be a separate function. It could be handled in the `handle_cast`

lib/nerves_hub_link/extensions/logging/logging.ex
```diff
@@ -0,0 +1,60 @@
...
+    {:noreply, state}
+  end
+
+  defp send_logs(log_payload, state) do
+    case push("send", log_payload) do
```
jjcarstens, line 55: I think we should have some sort of rate limiting on the client side as well. Otherwise we're going to bloat the connection which also controls updates

jjcarstens (commented):
Just some initial thoughts

lib/nerves_hub_link/extensions/logging/logging.ex
lawik: Maybe would be worthwhile to batch logs for ~500ms or something. Means we get natural batches on the insert side as well. Not exactly rate limiting but does improve the rate potentially.

More rate-limity: If a device is spewing logs it would make sense to shed some of them (it is probably looping/spewing the same thing continuously) and just include a log line about having shed logs?

lib/nerves_hub_link/extensions/logging/logging.ex
joshk: Good catch

lib/nerves_hub_link/extensions/logging/logger_handler.ex
joshk: Soooooo I added it, but its not available on Erlang versions under 27.x

Thoughts?

lib/nerves_hub_link/extensions/logging/logger_handler.ex
joshk: Hmmmmm, I hear you. I'll think about it. I like files for clarity and searchability.

lib/nerves_hub_link/extensions/logging/logging.ex
taun-fellowes: One incredibly out of date and maybe too complicated thought around the rate limiting topic is Mobius Telemetry. Maybe that is a separate extension? But in this context, if one treated a particular log event as telemetry, you could then log the event count and other metrics, rather than having to figure out how to limit a repetitive log entry. In other words, Mobius Telemetry naturally lends itself to rate limiting. And it has built in persistence so event counts aren't lost if there is limited connectivity. 

As an additional side note, Mobius already handles the health metrics used by the health extension.

taun-fellowes: Another thought in terms of client side implementation. We have implemented some custom Alerts. They are rare and unique. They do not fwap (mostly).  As in once they are set, they stay set until cleared. I am considering forwarding the "alarm set" as a log entry to the backend log extension. 

This should guarantee it being low volume. One benefit is there doesn't seem to be any alert history and alerts are polled very slowly (1 hr). And if an alert happens but clears within the hour, it is never noticed. We need these alerts. We could send them via MQTT to our backend, or use a 3rd party, or ?

nshoes (approved):

PR #312, opened 2025-06-26
Add a client callback for when the device is connected

amclain:
We needed a place to call `Nerves.Runtime.validate_firmware` when a connection is established to Nerves Hub, and it seemed like a callback in `Client` would be the best place for this. We did consider polling `NervesHubLink.connected?`, but this felt wrong since NervesHubLink has the `Client` handlers.

Although we're validating firmware in a custom handler, I implemented the NervesHubLink default handler as a no-op so it doesn't impact anyone. However, if it ever became the Nerves default to have [nerves_fw_autovalidate](https://github.com/nerves-project/nerves_system_bbb/blob/a774df9d83a667578cf2d2558a61d7c79cd642e2/uboot/uboot.env#L45-L49) disabled, this would be a convenient place for everyone's cloud-connected firmware to be validated by default. 😉

### DIFF
--- a/lib/nerves_hub_link/client.ex
+++ b/lib/nerves_hub_link/client.ex
@@ -70,6 +70,13 @@ defmodule NervesHubLink.Client do
           | {:progress, 0..100}
 
   @doc """
+  Called when the connection to NervesHub has been established.
+
+  The return value of this function is not checked.
+  """
+  @callback connected() :: any
+
+  @doc """
   Called to find out what to do when a firmware update is available.
 
   May return one of:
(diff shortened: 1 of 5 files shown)


joshk (approved):
Looks good to me.

@jjcarstens @nshoes @lawik ?

lawik (approved):
Makes sense to me :)

nshoes (approved):
LGTM! :) CI is throwing a fit, however.

joshk:
> LGTM! :)

That fit is spell check related :/ 
Looks like the deps need a new NPM version.

joshk:
Tried the close/reopen trick to run the tests again, no dice.

@amclain could you please rebase this against main, we have merged in a fix for the CI build. Thanks :)

joshk (changes requested):
Before merging this in, it occurred to me that we should make this optional, which also means we need to follow the pattern used by `reconnect_backoff` and `reboot`, such that `NervesHubLink.Client` defines a default to use.

joshk (approved):
Since this is an optional callback where the result isn't used, I'm happy with the implementation.

PR #318, opened 2025-08-12
Add opt-in support for websocket `permessage-deflate`

joshk:
Although we have set `compress: true` for the sockets defined in the `DeviceEndpoint`s in NervesHub, we also need to enable it in the client so it can be negotiated.

This PR adds opt-out support for compression, with compression on by default.

### DIFF
--- a/lib/nerves_hub_link/configurator.ex
+++ b/lib/nerves_hub_link/configurator.ex
@@ -36,6 +36,7 @@ defmodule NervesHubLink.Configurator do
     """
 
     defstruct archive_public_keys: [],
+              compress: false,
               connect: true,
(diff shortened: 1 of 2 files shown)


lawik:
Oooooh o.O

joshk:
@nshoes @jjcarstens what are your thoughts on enabling compression by default?

jjcarstens:
Can extensions be updated at runtime after starting the socket? If so, we could default to off and let the server tell us it supports deflation, then enable.

What happens if the device is deflating but server is not? Is that easy to bubble up as a problem?

jjcarstens (approved):
Perfect. On by default does not seem bad then

PR #321, opened 2025-08-14
Fixed the issue where default cacerts would override cacertfile option

D4no0:
The `LocalCertKey` configurator was always adding `cacerts`, even when `cacertfile` was provided.

The issue with that is the fact that `ssl` will ignore `cacertfile` option if `cacerts` is present, hence you will not be able to set that option at all.

### DIFF
--- a/lib/nerves_hub_link/configurators/local_cert_key.ex
+++ b/lib/nerves_hub_link/configurators/local_cert_key.ex
@@ -27,7 +27,14 @@ defmodule NervesHubLink.Configurator.LocalCertKey do
   end
 
   defp maybe_add_cacerts(%{ssl: ssl} = config) do
-    %{config | ssl: Keyword.put_new(ssl, :cacerts, Certificate.ca_certs())}
+    ssl =
+      if ssl[:cacerts] || ssl[:cacertfile] do
+        ssl
+      else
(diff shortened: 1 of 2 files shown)


lib/nerves_hub_link/configurators/local_cert_key.ex
```diff
@@ -27,7 +27,14 @@ defmodule NervesHubLink.Configurator.LocalCertKey do
...
   defp maybe_add_cacerts(%{ssl: ssl} = config) do
-    %{config | ssl: Keyword.put_new(ssl, :cacerts, Certificate.ca_certs())}
+    ssl =
+      if ssl[:cacerts] || ssl[:cacertfile] do
+        ssl
+      else
+        Keyword.put_new(ssl, :cacerts, Certificate.ca_certs())
+      end
+
+    %{config | ssl: ssl}
```
jjcarstens, line 37: Since we're already checking both keys, there would be no need to `Keyword.put_new` and we can directly replace. That could simplify just a bit, but not critical
```suggestion
    if ssl[:cacerts] || ssl[:cacertfile] do
      config
    else
      %{config | ssl: Keyword.put(ssl, :cacerts, Certificate.ca_certs())}
    end
```

PR #322, opened 2025-08-15
Add a simple adapter interface for settings and clearing alarms

joshk:
[`Alarmist`](https://hexdocs.pm/alarmist) is a cleaner and more advanced way of working with alarms.

While we recommend `Alarmist` is included in Nerves codebases, we should be able to fallback to `:alarm_handler`.

This PR adds a thin adapter-ish interface to using alarms within `NervesHubLink`, primarily supporting deduplication of alarms if `:alarm_handler` is being used.

I also added `:alarmist` to the deps list (optional) so Elixir starts the deps in the required order. Without this, `Code.ensure_loaded?(Alarmist)` would return false the first time its called.


lib/nerves_hub_link/alarms.ex
```diff
@@ -0,0 +1,41 @@
+# SPDX-FileCopyrightText: 2025 Josh Kalderimis
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+defmodule NervesHubLink.Alarms do
+  @moduledoc """
+  A slim adapter for `Alarmist` and `:alarm_handler`, providing a unified interface for setting and clearing alarms.
```
jjcarstens, line 7:
```suggestion
  A slim adapter for `Alarmist` and `:alarm_handler`, providing a unified interface for setting and clearing alarms.
  
  This primarily exists because `:alarm_handler.get_alarms()` will fail if the default alarm_handler
  has been replaced. Until a path is decided for account for that, this keeps the implementations
  separate with minimal set/clear handling adjustment when the default handler is used in order to
  match Alarmist experience better.
```

jjcarstens (approved):
Makes sense to me. I added a note to the module with the reasoning that might make it more clear. I believe the goal is we want Alarmist in the future, but that requires somehow helping determine `:alarm_handler.get_alarms` usage and warning that needs to be replaced with `Alarmist.get_alarms`. Until this, this is an easy path to get it used with devices

PR #327, opened 2025-08-24
A new firmware update workflow orchestrator

joshk:
This was inspired by #298 

With some simple changes to the `Downloader`, we can now stream chunks to a file. Couple this together with a set of `Updater`s, where different workflows associated with streaming straight to Fwup or from a file are encapsulated, we can now download firmware files to disk, which are recoverable between reboots, and then apply the updates once the download is complete.

I've given this a good testing with my dev RPI devices, and its working well.

Needs some docs, and some tests, and some tweaks on what statuses are reported and streamed to NH.

You can enable the new updater with:
```
config :nerves_hub_link,
  updater: NervesHubLink.UpdateManager.CachingUpdater
```

and you can configure where files are cached to using:
```
config :nerves_hub_link, CachingUpdater,
  cache_dir: "/data/my/dir/of/caches"
```
(the default is `/data/nerves_hub_link/firmware`)

You can also update the updater via iex using:
```
NervesHubLink.UpdateManager.change_updater(NervesHubLink.UpdateManager.StreamingUpdater)
```

### DIFF
--- a/lib/nerves_hub_link/application.ex
+++ b/lib/nerves_hub_link/application.ex
@@ -33,10 +33,15 @@ defmodule NervesHubLink.Application do
           update_available: &Client.update_available/1
         }
 
+        downloader_config = %{
+          cache_firmware_to_disk: config.cache_firmware_to_disk,
+          cache_firmware_dir: config.cache_firmware_dir
+        }
+
         [
           {DynamicSupervisor, name: ExtensionsSupervisor},
           Extensions,
-          {UpdateManager, fwup_config},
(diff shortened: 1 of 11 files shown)


nshoes (approved):
This LGTM. I tested:

- Updating multiple times with `StreamingUpdater`
- Updating after pulling the ethernet with `StreamingUpdater`
- Updating multiple times with `CachingUpdater`
- Updating after pulling the ethernet with `CachingUpdater`

Everything worked as expected!

PR #335, opened 2025-09-06
Add an optional `Client` callback for checking firmware validation

joshk:
This modifies the recently merged in #331 by delegating the firmware validation check to the device (via `Client`) instead of having NervesHub decide this.

The reason is two-fold:

1. having NervesHub decide if a firmware is valid requires all devices to implement the same logic in their KV data, which isn't the case.
2. and by using an optional callback on `Client` (with a default implementation) we can allow for developers to customize this logic.

This will require a change in NervesHub. A PR will follow shortly.

### DIFF
--- a/lib/nerves_hub_link/client.ex
+++ b/lib/nerves_hub_link/client.ex
@@ -46,6 +46,7 @@ defmodule NervesHubLink.Client do
   ```
   """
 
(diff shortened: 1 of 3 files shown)


nshoes (approved):
I like this! Sweet.

PR #337, opened 2025-09-09
Use a `Task.Supervisior` for executing support scripts

joshk:
- We can now monitor how many support scripts are running
- Plus, support scripts now have enforced timeouts
- And each support script can have its own timeout

### DIFF
--- a/lib/nerves_hub_link/support_scripts_manager.ex
+++ b/lib/nerves_hub_link/support_scripts_manager.ex
@@ -9,9 +9,17 @@ defmodule NervesHubLink.SupportScriptsManager do
   """
   use GenServer
 
+  alias NervesHubLink.Socket
+
   require Logger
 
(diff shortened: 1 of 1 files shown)


nshoes:
@joshk ETA on the hub changes? I'd like to test this, but I'd rather wait until both systems are communicating properly.

joshk:
> @joshk ETA on the hub changes? I'd like to test this, but I'd rather wait until both systems are communicating properly.

@nshoes I've just updated the PR so the result is backwards compatible, so it should be fine for you to test

joshk:
@nshoes @jjcarstens thoughts on this PR?

nshoes:
This PR slipped past me... I've got this in my todos for today or tomorrow.

joshk:
Pinging @nshoes and @lawik

PR #342, opened 2025-09-27
Support a richer set of firmware update progress messages

joshk:
_Please note: This PR builds upon the new Updater Workflow which is yet to be merged in._

This PR enriches the firmware update status messages to better communicate the stage of the update workflow.

The happy path includes:

- firmware update request received
- update file downloading
- applying/updating system using the update file
- update completed

But we also now send messages for:

**Firmware update is ignored**

This is a new message, which allows hub to clear away the inflight update and take additional actions like disabling automatic updates or blocking updates for x period of time. A reason why the update was ignored can also be included.

**Firmware update is rescheduled**

Instead of having the device take care of rescheduling the update, we now tell Hub to do this for us. The rescheduling is done in minutes, not milliseconds, with a minimum of 5 minutes. An optional reason can also be included. This works perfectly with the Deployment Orchestrator, but we will need to think about how to deal with manually requested updates, as there isn't a process or mechanism we can hook into right now.

**Firmware failed**

If an update fails, for whatever reason, we don't currently communicate this with Hub, which means a device will take an 'update slot' (Deployment Groups can limit how many devices are updated at once) until it is eventually cleared away. This message allows us to track failures and clear away `InflightUpdate`s.

---

These changes are backwards compatible with current and older Hub versions, as the new information uses current PubSub events with some extra enrichment. 

Once we lock in and agree on the new events and information, I can start working on the Hub side of things.


fhunleth:
I don't have deep comments. I'm quite happy that firmware update rescheduling is going to the server side. Letting the device request an upgrade is fine, but just postponing a request indefinitely and then acting on it always seemed so risky. Glad it is en route to deletion. The rest of what you said seemed really good too.

nshoes (approved):
Love this! I ran through the happy path on a device and everything looks good. Looking forward to the hub-side implementation 😈

PR #344, opened 2025-09-29
Use `:gen_tcp.connect/4` as it supports specifying a timeout

joshk:
`:inet.gethostbyname/1` doesn't support specifying a timeout, and is known for sometimes taking 5+ seconds, which in turn blocks the Socket GenServer from responding to other messages, like `connected?/0`.

This change enables us to specify a timeout, resolving the issue mentioned above.

### DIFF
--- a/lib/nerves_hub_link/socket.ex
+++ b/lib/nerves_hub_link/socket.ex
@@ -468,15 +468,16 @@ defmodule NervesHubLink.Socket do
 
   @impl Slipstream
   def handle_info(:connect_check_network_availability, socket) do
-    hostname = URI.parse(socket.assigns.config.socket[:url]).host
+    uri = URI.parse(socket.assigns.config.socket[:url])
 
-    case :inet.gethostbyname(to_charlist(hostname)) do
-      {:ok, _} ->
+    case :gen_tcp.connect(to_charlist(uri.host), uri.port, [active: false, packet: 0], 2_000) do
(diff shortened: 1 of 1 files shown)


lib/nerves_hub_link/socket.ex
```diff
@@ -468,15 +468,16 @@ defmodule NervesHubLink.Socket do
...
   @impl Slipstream
   def handle_info(:connect_check_network_availability, socket) do
-    hostname = URI.parse(socket.assigns.config.socket[:url]).host
+    uri = URI.parse(socket.assigns.config.socket[:url])
 
-    case :inet.gethostbyname(to_charlist(hostname)) do
-      {:ok, _} ->
+    case :gen_tcp.connect(to_charlist(uri.host), uri.port, [active: false, packet: 0], 2_000) do
+      {:ok, tcp_socket} ->
+        :gen_tcp.close(tcp_socket)
```
nshoes, line 475: Any considerations needed since we're opening and closing a socket now? The previous implementation just resolved the hostname.

joshk:
Resolving a domain name is using a socket too, it’s just doing some other
stuff as well.

There is :inet_res but that’s only available on erlang 28.1

This is a better solution because it removes the risk of blocking the
genserver.

nshoes:
@joshk Sweet, sounds good.

nshoes (approved):

PR #349, opened 2025-10-13
Add a 'local shell' extension for establishing an interactive shell

joshk:
Inspired by https://github.com/nerves-containers/nerves_ssh_system_shell

Screenshot 2025-10-13 at 5 24 31 PM

Built with the idea in mind that shell sessions can be opened and closed, which is nice for possible auditing purposes.

(Hub PR to come after this PR is reviewed)


joshk:
> Good move on the optional dependency. Would it matter to make the extension disabled if the dep isn't there? Might not matter.
> 
> This is awesome :)

@lawik Great idea, just pushed a change to make this conditional, would you mind checking it and giving me your feedback?

PR #351, opened 2025-10-18
Add support for refreshing the sockets config

joshk:
The notable changes include:

- `NervesHubLink.refresh_config()` can now be used to refresh the config used by the underlying connection. Coupled with `Client.ready_to_connect?()`, you can provision a TPM or NervesKey, refresh the config, and then signal to the socket that it can connect.

- `Client.ready_to_connect?()` can be used to delay the socket's connection until certain conditions are met, eg, wait for a device's TPM or NervesKey to be initialized. If `ready_to_connect?/0` returns false, it will be checked periodically using a 5 second delay.

- If the application config of `connect: false` is used, `NervesHubLink.establish_connection()` can be called to manually establish the connection.

- And `NervesHubLink.disconnect!()` can be used to disconnect the socket and disable automatic reconnection.

A byproduct of these changes is that the Application supervisor children are always started, which allows for manual connection management.

### DIFF
--- a/lib/nerves_hub_link/client.ex
+++ b/lib/nerves_hub_link/client.ex
@@ -110,7 +110,7 @@ defmodule NervesHubLink.Client do
 
   The default implementation returns `true`.
   """
-  @callback ready_to_connect?() :: bool()
+  @callback ready_to_connect?() :: boolean()
 
   @doc """
   Called when the connection to NervesHub has been established.


lib/nerves_hub_link/socket.ex
```diff
@@ -133,11 +148,20 @@ defmodule NervesHubLink.Socket do
...
-      {:ok, socket, {:continue, :connect}}
+    cond do
+      not config.connect ->
+        {:ok, socket}
+
+      not Client.ready_to_connect?() ->
+        schedule_establish_connection_check()
+        {:ok, socket}
+
+      config.connect_wait_for_network ->
```
jjcarstens: Is the assumption here that if you ever implement checks to return `false` on `ready_to_connect?`, you will skip any `wait_for_network` check?  If I've traced things right, that seems to be the behavior

joshk: That was the original intent, but I can change this, thoughts?

PR #352, opened 2025-10-23
Add logging that helps indicate clock issues

lawik:
We had a bunch of people dealing with time making shared secrets not work for the Goatmire badge. I have sometimes thought I had that problem when my clock was correct and there were other issues on the backend. This should make it easier to know what time the device is working with when creating the auth header.

### DIFF
--- a/lib/nerves_hub_link/configurators/shared_secret.ex
+++ b/lib/nerves_hub_link/configurators/shared_secret.ex
@@ -46,7 +46,7 @@ defmodule NervesHubLink.Configurator.SharedSecret do
       |> Keyword.put(:signed_at, System.os_time(:second))
 
     datetime = DateTime.from_unix!(opts[:signed_at]) |> DateTime.to_iso8601()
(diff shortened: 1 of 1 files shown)


lib/nerves_hub_link/configurators/shared_secret.ex
```diff
@@ -43,6 +45,9 @@ defmodule NervesHubLink.Configurator.SharedSecret do
       # quickly. See https://www.erlang.org/doc/apps/erts/time_correction#erlang-system-time
       |> Keyword.put(:signed_at, System.os_time(:second))
 
+    datetime = DateTime.from_unix!(opts[:signed_at]) |> DateTime.to_iso8601()
+    Logger.info("Generating auth headers with time #{datetime}")
```
joshk:
```suggestion
    Logger.info("[NervesHubLink:SharedSecret] Generating auth headers with time #{datetime}")
```

PR #359, opened 2025-10-25
Various small improvements I've come across

joshk:
Not really big enough for their own PRs

- Fix a warning when generating docs
- Add a debug log line when the extensions channel connects
- Remove some old commented out code
- And remove some unused code


lib/nerves_hub_link/socket.ex
```diff
@@ -726,22 +714,6 @@ defmodule NervesHubLink.Socket do
   defp maybe_cancel_timer(nil), do: :ok
   defp maybe_cancel_timer(pid), do: Process.cancel_timer(pid)
 
-  defp handle_join_reply(%{"firmware_url" => url} = update, socket) when is_binary(url) do
-    case UpdateInfo.parse(update) do
-      {:ok, %UpdateInfo{} = info} ->
-        UpdateManager.apply_update(info, socket.assigns.config.fwup_public_keys)
```
lawik: Are updates never sent in the join reply anymore?

PR #360, opened 2025-10-26
Add foundational support for multiple Sockets to be started

joshk:
A continuation of #127


joshk:
Same issue as Connor mentioned, there already exists a reconnect function
and defining a new one makes elixir not happy.

We can call it something else, suggestions welcome. Since Socket isn’t
publicly documented we can use whatever name we want.

PR #361, opened 2025-10-26
Move to GitHub Actions for CI

joshk:
This CI setup is feature comparable with the previous Circle setup, with the addition of Dialyzer.

One of the main reasons for the move is because I don't have access to manage the Circle settings, including the ability to do things like trigger rebuilds. Plus, with NervesHub using GitHub Actions, it seemed logical to make them both use the same setup.

The only thing I don't love about this CI setup is that installing the debs and using `apt install` would sometimes lead to 1-2min slow steps, while the rest of the job completes in 10 seconds.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,213 @@
...
+          cache-name: cache-compiled-build
+        with:
+          path: _build
+          key: ${{ runner.os }}-${{ steps.beam.outputs.elixir-version }}-${{ steps.beam.outputs.otp-version }}-mix-${{ env.cache-name }}-${{ hashFiles('**/mix.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ steps.beam.outputs.elixir-version }}-${{ steps.beam.outputs.otp-version }}-mix-${{ env.cache-name }}-
+            ${{ runner.os }}-${{ steps.beam.outputs.elixir-version }}-${{ steps.beam.outputs.otp-version }}-mix-
+
+      - name: Clean to rule out incremental build as a source of flakiness
+        if: github.run_attempt != '1'
```
jjcarstens: 🎉 This is a neat trick

.github/workflows/ci.yml
```diff
@@ -0,0 +1,213 @@
...
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Elixir
+        id: beam
+        uses: erlef/setup-beam@v1
+        with:
+          otp-version: 27.3.4
+          elixir-version: 1.18.4
```
jjcarstens:
```suggestion
          version-file: .tool-versions
          version-type: strict
```
I think we should rely on the version specified here rather than hard code

jjcarstens (approved):

.circleci/config.yml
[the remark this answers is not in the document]
joshk: I'll do that once Circle is disabled. If Circle can't find the config it would cause builds to error.

.github/workflows/ci.yml
joshk: Ha, funny, I didn't even think of that because I forgot we have it.

joshk: Wait, we don't have one! Guess I'll need to add it.

PR #387, opened 2026-01-08
v2.10.0

lawik:


lawik:
Frank requested a new release to get rid of some Elixir 1.19 warnings. We are overdue for a release I think.

@joshk local shell should be fine to ship right even if it isn't used by default?

PR #400, opened 2026-02-11
Report network interface after connecting to NervesHub

nshoes:
This PR has two main goals:

1.) Report network interface to NervesHub after connecting
2.) When starting a firmware update, report if there's a mistmatch between the interface that `Socket` and `Downloader` uses. This only for monitoring purposes NervesHub-side. We want to be confident that the interface is the same, but are unsure of what really happens. The data will tell.

Getting the interface requires some PID spelunking and we want to keep `Socket` safe at all costs, so it's wrapped in a `rescue`. Starting a firmware download gets the same treatment.

Todo:
- [x] Derive "download network interface" from Mint's socket, not `Socket`'s socket, when downloading
- [x] Make the comparison, report mismatch

### DIFF
--- a/lib/nerves_hub_link/socket.ex
+++ b/lib/nerves_hub_link/socket.ex
@@ -616,8 +616,8 @@ defmodule NervesHubLink.Socket do
   def handle_info(:get_network_interface, socket) do
     with pid when is_pid(pid) <- Slipstream.Socket.channel_pid(socket),
          %{conn: conn} <- :sys.get_state(pid),
-         socket = Mint.HTTP.get_socket(conn),
-         interface when is_binary(interface) <- NetworkInterface.from_socket(socket) do
+         underlying_socket = Mint.HTTP.get_socket(conn),
+         interface when is_binary(interface) <- NetworkInterface.from_socket(underlying_socket) do
       _ = report_network_interface(socket, interface)
       {:noreply, assign(socket, network_interface: interface)}
     else
@@ -805,6 +805,6 @@ defmodule NervesHubLink.Socket do
   end
 
   defp report_network_interface(socket, interface) do
-    push(socket, @device_topic, "set_network_interface", %{interface: interface})
+    push(socket, @device_topic, "report_network_interface", %{interface: interface})
   end
 end
--- a/lib/nerves_hub_link/network_interface.ex
+++ b/lib/nerves_hub_link/network_interface.ex
@@ -15,15 +15,7 @@ defmodule NervesHubLink.NetworkInterface do
   """
   @spec from_socket(:ssl.socket() | :inet.socket() | Mint.Types.socket()) :: nil | binary()
   def from_socket(socket) do
-    case socket do
-      {:sslsocket, _, _} ->
-        {:ok, {address, _}} = :ssl.sockname(socket)
-        address
-
-      _ ->
-        {:ok, {address, _}} = :inet.sockname(socket)
-        address
-    end
+    address_from_socket(socket)
     |> interface_from_address()
   rescue
     err ->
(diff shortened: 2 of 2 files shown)


lawik:
Is the reason to not find and send this information immediately that it is slow or something? My thinking is it would make sense to report it directly on join if we can. And that should be possible since the socket is already established when we make the join if I recall correctly.

nshoes:
> Is this reported for display in NervesHub

No, that's specifically to see if downloading with a different interface is ever an issue. We have an inkling that it might be, so we're reporting and just monitoring. I hope that it won't generally won't happen, but the data will tell the real story.

> My thinking is it would make sense to report it directly on join if we can. And that should be possible since the socket is already established when we make the join if I recall correctly.

I tried this, and it never works. Assuming it's timing. We need the channel pid to get to the underlying socket layer, and it's not available until we make the connection to NH. Let me know if you have any ideas, I really wanted to send this info as a join param.

lib/nerves_hub_link/downloader.ex
```diff
@@ -453,6 +454,8 @@ defmodule NervesHubLink.Downloader do
              transport_opts: transport_opts
            ),
          {:ok, conn, request_ref} <- Mint.HTTP.request(conn, "GET", path, request_headers, nil) do
+      :ok = maybe_report_network_interface_mismatch(conn)
```
joshk: I was thinking, instead of sending such a specific message of 'mismatch', how about we add the interface/connection type to the 'download started' message?

joshk (commented):
I love where this is going.

Two suggestions I have:

1. instead of sending the interface and pattern match in Hub (eg. `wwlan*` -> `lte`), lets do that in Link and define some richer connection types, eg. `ethernet`, `wifi`, `lte` (general purpose cellular), `lte-cat-1, `lte-cat-4`, `lte-cat-nb1`, `lte-cat-nb2`, `5g`) in Hub which Link can send up. This would allow us to enhance this code (at a later date) to support interface -> link type matching.

2. for this PR or a later PR, we can use the `ifaddr` option when establishing the download to restrict the download to the same interface as the websocket connection.

lib/nerves_hub_link/socket.ex
```diff
@@ -193,7 +202,10 @@ defmodule NervesHubLink.Socket do
   @impl Slipstream
   def handle_join(@device_topic, _reply, socket) do
     Logger.debug("[#{inspect(__MODULE__)}] Joined Device channel")
-    {:ok, assign(socket, joined_at: System.monotonic_time(:millisecond))}
+
+    interface = maybe_report_network_interface(socket)
+
+    {:ok, assign(socket, joined_at: System.monotonic_time(:millisecond), interface: interface)}
```
jjcarstens:
```suggestion
    {:ok, assign(socket, joined_at: System.monotonic_time(:millisecond), network_interface: interface)}
```
This is mismatched from what is being used in the assigns elsewhere

nshoes: Thanks!

lib/nerves_hub_link/socket.ex
```diff
@@ -606,6 +621,22 @@ defmodule NervesHubLink.Socket do
     {:noreply, stop_iex(socket)}
   end
 
+  def handle_info({:get_network_interface, attempts: attempts}, socket) when attempts >= 10 do
```
jjcarstens: I don't think you need to cap attempts. I suggest letting it go until discovered. That way you don't have to deal with cases that it fails 10 times, but succeeds on the 11th. If it never gets discovered, that would be a glaring problem to fix.

PR #412, opened 2026-04-09
Send :status message to UpdateManager by default

nshoes:
This message isn't handled by `Socket` anymore, it's handled by [`UpdateManager`](https://github.com/nerves-hub/nerves_hub_link/blob/fix-default-arg-status/lib/nerves_hub_link/update_manager.ex#L62), so the default should reflect.


kybishop (approved):

joshk:
@nshoes do we use that anywhere? I'm wondering if we should rename it from `status/1` to `updater_status/1`?

PR #418, opened 2026-05-05
Add erlang socket address resolution to network interface

### LINKED ISSUE #417: Network insterface reporter Erlang socket warning
I kept getting this error 
- `Error: could not retrieve network interface: %FunctionClauseError{module: :prim_inet, function: :sockname, arity: 1, kind: nil, args: nil, clauses: nil}`

And found that `address_from_socket` in `NervesHubLink.NetworkInterface` did not support eralng socket refrences.
There might be a reason for this that I am unaware of, but for my own sanity I added this to my fork to fix it.
```Elixir
# lib/nerves_hub_link/network_interface.ex:38
defp address_from_socket(socket) when is_reference(socket) do
  {:ok, %{addr: address}} = :socket.sockname(socket)
  address
 end
```
If this is correct, should I make a PR?

OleHenFo:
Fixes #417
Adding a function header to match erlang sockets in netwrok interface.

### DIFF
--- a/lib/nerves_hub_link/network_interface.ex
+++ b/lib/nerves_hub_link/network_interface.ex
@@ -14,7 +14,8 @@ defmodule NervesHubLink.NetworkInterface do
   Get the network interface from a Slipstream.Socket or Mint.HTTP struct. This is used to report the
   network interface being used for the connection to NervesHub and for firmware downloads.
   """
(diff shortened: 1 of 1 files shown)


joshk (approved):
Awesome! Thank you!

PR #425, opened 2026-05-11
Add mint http opts passthrough

lawik:


joshk:
Nice addition! Once you have tested it with a read life proxy, we can merge it in.

lawik:
Confirmed :)

PR #433, opened 2026-05-28
Network interface reporting improvements

kybishop:
This PR was started to deal with a breakage between OTP 27 -> OTP 28. I stepped back to see if there was a more stable way of determining network interface, and landed on this UDP socket approach.

1. No data is ever sent to the server
2. This should be more stable between upgrades, both in elixir/otp and deps like slipstream and mint
3. We now only send network interface data if it changed between the last attempt, which should help with flapping devices

This approach has been tested on my device running OTP 28 and the latest nerves_system_br

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -21,7 +21,7 @@
     "youngblood",
 
     // Technical terms
-    "nolink",
+    "Geolocation",
     "backoffs",
     "bingenerate",
     "configurator",
@@ -29,12 +29,13 @@
     "disksup",
     "expty",
     "fwupdate",
-    "Geolocation",
     "geoip",
+    "getaddr",
     "getcap",
     "kbps",
     "maint",
     "nerveshub",
+    "nolink",
     "nvread",
     "privkey",
     "reconnections",
--- a/lib/nerves_hub_link/downloader.ex
+++ b/lib/nerves_hub_link/downloader.ex
@@ -511,7 +511,7 @@ defmodule NervesHubLink.Downloader do
     with {:ok, conn} <-
            Mint.HTTP.connect(String.to_existing_atom(scheme), host, port, connect_opts),
          {:ok, conn, request_ref} <- Mint.HTTP.request(conn, "GET", path, request_headers, nil),
-         :ok <- report_download_started(conn) do
+         :ok <- NervesHubLink.send_update_status({:started, NetworkInterface.from_uri(uri)}) do
       {:ok,
        %Downloader{
          state
@@ -568,11 +568,6 @@ defmodule NervesHubLink.Downloader do
   defp close_conn(%Downloader{conn: conn}) do
     {:ok, _} = Mint.HTTP.close(conn)
     :ok
-  end
-
-  defp report_download_started(conn) do
-    downloader_network_interface = NetworkInterface.from_socket(Mint.HTTP.get_socket(conn))
-    NervesHubLink.send_update_status({:started, downloader_network_interface})
   end
 
   # Shallow-merge user-supplied opts on top of the base, but for :transport_opts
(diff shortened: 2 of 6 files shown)


jjcarstens (commented):
What exactly was the problem between OTP 27 -> 28?

I don't think this will quite work. When you use a host with UDP, you're not testing the _actual_ socket that is connected. It will use whatever primary interface is available which may be different than the actual socket.

Consider a case like this:
`Device Boot -> Connects to LTE -> NHL makes connection -> WIFi connects -> NHL Checks network interface with UDP -> UDP reports wlan0`

The NHL connection will be over the cellular `wwan0` interface since that was available _before_ WiFi and while NHL was connecting, but WiFi finishes connecting before it finishes checking the network interface. In that case, you would get `wlan0`.

I've simulated this on device:
```elixir
iex(7|macaw)> VintageNet.deconfigure("wlan0")
:ok

iex(8|macaw)> NervesHubLink.reconnect()
:ok

iex(9|macaw)> test_interfaces.()
%{current: "wwan0", ky_udp: "wwan0"}

iex(10|macaw)> VintageNet.reset_to_defaults "wlan0"
:ok

iex(11|macaw)> test_interfaces.()
%{current: "wwan0", ky_udp: "wlan0"}
```

<details>
<summary>test_interfaces code</summary>

Renamed the `NervesHubLink.NetworkInterface` in this PR to `Ky.NetworkInterface`, pasted into IEx. Then ¬

```elixir
test_interfaces = fn ->
  slipstream = :sys.get_state(NervesHubLink.Socket)
  p = Slipstream.Socket.channel_pid(slipstream)
  socket = Mint.HTTP.get_socket(:sys.get_state(p).conn)
  %{
    current: NervesHubLink.NetworkInterface.from_socket(socket),
    ky_udp: Ky.NetworkInterface.from_uri(slipstream.assigns.config.socket[:url])
  }
end
```
</details>

This is because once a socket is established, it will stay on the interface used and typically not hop to another one unless explicitly told to do so. This is why the original implementation went through effort to get the _actual_ socket shortly after it connected and see which interface it used rather than try to create _another_ socket for testing and assuming they both would be the same.

kybishop:
I'm actually more surprised to learn we don't reconnect the socket when a better interface becomes available 😅 

I'll think on this. Perhaps we can do this:

1. Fix things on otp 28
2. Implement (rate-limited?) reconnections when higher priority interfaces become available
3. Switch to this approach so we don't have to rely on the internals of Slipstream and Mint.

jjcarstens:
It's mostly by design. The intent is to always favor a connection to the update server. The more clever you try to get, the more you risk that connection. So it is left to other layers to decide to make that decision or not since NHL does not have preference on how things connect (it's usually VintageNet, but could be other ways technically)

I would vote for fixing OTP 28 and not attempting to get clever with reconnects. Then if we want to avoid Slipstream/Mint internals, we upstream helper functions to them or add what is needed to be less internal. (Which isn't really that internal since they mostly give us functions to find what we need)

Though I think this is an acceptable case for now to use internals with appropriate tests and a matrix of OTP versions tested in CI which would catch it. It's not really a common occurrence to break at this level, and when it does it is fairly simple to catch and adjust

kybishop:
> It's mostly by design. The intent is to always favor a connection to the update server. The more clever you try to get, the more you risk that connection. So it is left to other layers to decide to make that decision or not since NHL does not have preference on how things connect (it's usually VintageNet, but could be other ways technically)
> 
> I would vote for fixing OTP 28 and not attempting to get clever with reconnects. Then if we want to avoid Slipstream/Mint internals, we upstream helper functions to them or add what is needed to be less internal. (Which isn't really that internal since they mostly give us functions to find what we need)
> 
> Though I think this is an acceptable case for now to use internals with appropriate tests and a matrix of OTP versions tested in CI which would catch it. It's not really a common occurrence to break at this level, and when it does it is fairly simple to catch and adjust

Makes sense! Closing out to go with that approach

PR #434, opened 2026-05-28
Recover SharedSecret signing when the device clock is wrong

lawik:
SharedSecret auth signs with the device clock and assumes it's within ~90s of the server. New devices often boot with a clock far off reality and can take a long time to recover, blocking the first connection.

When a connection fails, pick up the `Date` header from the response and use it as the signing time on the next attempt. If the device clock has since caught up past the hint, the device clock wins.

A nonce exchange would be cleaner but needs server cooperation; this is a backwards compatible device side workaround.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -38,6 +38,7 @@
     "nvread",
     "privkey",
     "reconnections",
+    "strftime",
     "subkey",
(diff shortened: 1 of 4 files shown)


joshk (approved):
smart!

PR #457, opened 2026-08-15
Add the external identity extension

joshk:
The client half of nerves-hub/nerves_hub_web#2896.

Lets a device report the identities it holds on networks NervesHub doesn't run — an iroh endpoint id, a NetBird or Tailscale peer key, a WireGuard public key — so an operator can see them on the device page and reach the device without going through NervesHub.

## No default provider, on purpose

NervesHubLink has no way to know that a device is also an iroh endpoint. The library that owns that identity does, so this takes a pluggable provider, following the `Geo.Resolver` precedent:

```elixir
config :nerves_hub_link,
  external_identity: [providers: [MyApp.IrohIdentity]]
```

Writing one is a single `identity/0` callback. The guide has a working `iroh_console` example and a NetBird sketch.

Built-in providers are not included for the supported services. WireGuard's key is in VintageNet config, iroh's is in a running endpoint — so scraping `wg show` or `netbird status --json` is the wrong instinct for this platform, and depending on `iroh_beam` (a precompiled Rust NIF) in the library every device runs is a bad trade for the fraction using it. The right home for a provider is the library that owns the identity; `iroh_console` ships one in a companion PR.

## Instances

A device can run two iroh endpoints — this library's console, plus whatever the application uses iroh for — each with its own key. Both are `iroh`, told apart by an optional `instance`. Omit it and NervesHub treats it as the service's only endpoint. It must be stable: the server keys the identity on it, so an instance derived from the key itself would create a new record on every rotation.

## Failure behaviour

The extension is non-critical and must never disturb the socket, so:

- `:unavailable` is logged at **debug**, not warning. A unit that legitimately isn't on a given network reconnects like any other, and warning each time would make the warnings worthless. `{:error, reason}` is the warning case.
- A provider that raises is caught and skipped — one broken provider never costs the others their identities.
- Being asked with no providers configured logs a warning, since the server only asks when an operator switched it on, which makes an empty provider list a likely half-finished setup.

There's no interval; an identity is long-lived, so the server asks once per connection. `send_report/0` announces a mid-connection change, and `identities/0` shows what the providers currently return without sending anything.

Added to `@default_extension_modules`, so upgrading picks it up — no behaviour change for anyone, since the server won't attach it unless the product opts in.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


joshk:
Folded into #460 along with #458. The extension arrives already named `network_identity` there, so a reviewer doesn't read the old name and then watch it get renamed.

PR #458, opened 2026-08-17
Rename external identity to network identity

joshk:
Every service this reports a key for is a network: iroh, NetBird, Tailscale, WireGuard. "External" was carrying only the information that the network is somebody else's, which is the least interesting thing about it.


joshk:
Folded into #460 along with #457, so the extension and its rename read as one change rather than a stack with a rename in the middle. Same tree, plus the polling work.

PR #463, opened 2026-08-21
Version 2.13.0

joshk:
Prepares the 2.13.0 release: rebased on `main`, folds the changelog entries the merged PRs wrote into a single `2.13.0` section, adds entries for the work that landed without them, and bumps `@version` in `mix.exs`.

## Highlights

**Added**
* Device-managed updates (#473) and the `PartsUpdater` download strategy (#462)
* Error Reports (#474) and Logging (#303, #470) extensions
* Network Identity extension (#460) and Msgpack serialization (#443)
* Opt-in manual supervision via `NervesHubLink.Supervisor` (#467)
* Shared Secret signing recovers from a wrong device clock; `Nerves.Runtime.firmware_slots/0` support (#442)

**Changed**
* Firmware and archives that cannot be signature verified are now refused rather than applied (#464)
* Extensions are negotiated with the server rather than offered blind (#468)
* Minimum Elixir is now 1.15; CI covers up to 1.20.3 on OTP 29 (#432, #456)

**Fixed**
* Downloader wrote signed URLs to the log, and resumed downloads could corrupt firmware (#462)
* `CachingUpdater` cache clearing (#461) and `cache_dir` config (#452)
* Compile-time `DeviceStatus` timestamp that broke reproducible builds (#451)

Every entry is one or two sentences, with the detail behind a link to the guide that covers it. That rewrites the entries the merged PRs wrote for their own `Unreleased` section, so it is worth a look if you would rather keep any of that wording.

Worth a second opinion before tagging: #464 means a device with no `fwup_public_keys` refuses firmware it previously installed unverified. That is a behaviour change for anyone running unsigned firmware, so 3.0.0 is defensible if you would rather not ship it in a minor.


joshk:
I think it's time for another release. I'm going to do some testing before I cut it (and pending your approval)

PR #467, opened 2026-08-22
Allow manual NervesHubLink supervision

elielgordondensity:
## Summary
Rationale - we have devices behind a proxy and proxy configs can change, which means reconfiguring link but the current setup does not make that easy or elegant. We could expose a reconnect function and that would help, but I think allowing folks to explicitly start and supervise link provides more flexibly and allows us to reason about failure in terms of our application structure.    

- add a reusable `NervesHubLink.Supervisor` for the connection process tree
- let applications opt out of automatic startup with `config :nerves_hub_link, start: :manual`
- document manual supervision and add focused supervisor wiring tests

## Testing
- `ASDF_ERLANG_VERSION=27.3.4.2 mix test test/nerves_hub_link/supervisor_test.exs`

Some readings that influenced this:

https://jola.dev/posts/let-libraries-be-libraries

https://keathley.io/blog/reusable-libraries.html


guides/configuration.md
```diff
@@ -1,5 +1,45 @@
 # Configuration
 
+## Supervision
+
+By default `:nerves_hub_link` is an OTP application that starts
+`NervesHubLink.Supervisor` automatically and builds its configuration from the
+application environment (i.e. your `config.exs`). This is the historical
+behaviour and requires no changes.
+
+If you'd prefer to own the connection's lifecycle — for example to start it
```
lawik:
```suggestion
If you'd prefer to own the connection's lifecycle - for example to start it
```

lawik (approved):

lawik:
@joshk looks good to me, any notes?

This makes custom flows for bringing up the networking and such way, way more ergonomic.

It may also means we can make a lot of tests more isolated.

nshoes (approved):
This is very cool! LGTM

lawik:
Spellchecking lint needs the word eliel ;)

joshk:
I've review shortly.

joshk:
I have a question around `config :nerves_hub_link, start: :manual`, are there pros or cons to using the built in `runtime: false` on the `nerves_hub_link` dependency itself, which would tell Elixir to include the dep but not start the Application.

elielgordondensity:
@joshk From my understanding runtime false would prevent otp from starting our otp extra applications so we would have to make sure we start them when we bringup the new supervisor. Usually library applications like Finch don't declare application module or support an auto start like we do and instead expose a supervisor which makes this easier. I'd let anyone else chime in if they have any experience here.

lawik:
Many Nerves libraries predate the practice and also in many cases using Erlang's applications mechanism make sense. But allowing more granular starting of nerves_hub_link is just good and useful.

A nice next step would be to make it not rely on module names for messaging genservers and then we'd be able to run multiple, which would be nice in tests.

I think this is a good change even though the mechanism could be done with runtime: false or connect: false along with Application.stop/ensure_all_started

joshk:
I tried to amend the PR so I could merge it but edits were blocked.

elielgordondensity:
@joshk its because I opened from our company fork. I can reopen from my personal or suggest changes? I think @lawik did that and it worked

joshk:
@elielgordondensity or if you fix those final spellcheck issues I can merge this in

elielgordondensity:
@joshk Added Eliel so satisfy cspell

joshk:
@elielgordondensity sorry, you have some test failures with the latest Elixir and Erlang

elielgordondensity:
@joshk Seems to be unrelated to my changes: The downloader test times out, increased it from 100 to 1000ms

joshk:
> @joshk Seems to be unrelated to my changes: The downloader test times out, increased it from 100 to 1000ms

Ahhhh, what I saw were compile warnings. Could you check for any compile warnings with the latest Elixir please.

elielgordondensity:
@joshk Took a look and only saw some xref stuff from x509 dep? Perhaps dependency caching suppressed those before? Doesn't seem like a path I touched

PR #473, opened 2026-08-29
Device-managed firmware updates

joshk:
Lets a device choose *when* it takes firmware. Its deployment group still chooses *what*.

The client half of nerves-hub/nerves_hub_web#2978 — that PR carries the protocol document and the server rules. **This needs a NervesHub that supports them**; against an older server nothing changes and the device keeps taking updates exactly as before.

## The API

Five functions, and no state kept on the device:

```elixir
NervesHubLink.update_mode()              #=> :off | :automatic | :device_managed | nil
NervesHubLink.managed_updates_allowed?() #=> may this device choose?
NervesHubLink.set_update_mode(:device_managed)
NervesHubLink.check_for_update()         #=> {:ok, %{available?: true, firmware_meta: %{..}}}
NervesHubLink.start_update()             #=> :ok | {:error, {:busy, 5}}
```

The mode is held by NervesHub rather than here. That is what makes a user-facing "automatic updates" switch workable — it survives a reboot without this library persisting anything, an operator can see and change it, and `update_mode/1` always reports what NervesHub actually has rather than what this device last wanted.

It returns `nil` until NervesHub says, which a server too old to know about update modes never will. Treat `nil` as "not device managed", never as an error.

## Checking and requesting are separate

`check_for_update/1` is allowed in every mode, `:off` included — it is read-only, so a frozen device can still tell its user an update exists and an administrator needs to act. It carries **no firmware URL**.

That is the point of the split. Firmware URLs are signed and time limited, so a device that checked at 02:00 and updated at 04:00 would find a dead link. `start_update/1` fetches one at the moment it is going to be used, and NervesHub answers with the same `update` message a deployment group sends — so nothing new applies firmware, and everything after the URL arrives is code that already shipped.

## Being told to wait

Deployment groups pace their rollouts, and a device asking for firmware takes a slot in that pacing like any other update — a fleet that all wakes at 03:00 local would otherwise walk straight past it.

```elixir
case NervesHubLink.start_update() do
  :ok -> :updating
  {:error, {:busy, minutes}} -> reschedule_in(minutes)
  {:error, reason} -> Logger.warning("update refused: #{inspect(reason)}")
end
```

`start_update/1` also refuses locally with `{:error, :updating}` while an update is already running, rather than spending a deployment slot on a firmware URL `UpdateManager` would ignore.

## What a device may not do

A device can move itself between `:automatic` and `:device_managed`, and only once an operator has allowed it — see `managed_updates_allowed?/1`, which is off until they do.

It can never set `:off`. That would let a device stop its own updates and leave nobody able to fix it remotely, so it stays an operator's decision.

## Implementation notes

Each request is answered by a **separate message** rather than a reply to the push, so the caller is parked in `pending_update_requests` until that message arrives, with a timeout so a lost answer cannot leak a stuck caller. `set_update_mode/2` is always answered with the mode NervesHub actually has — refusal included — so the two can never end up disagreeing about whether the device is being pushed to.

Error reasons from the server are **mapped, not converted**, so a reason NervesHub invents later cannot grow a device's atom table. An unknown reason still resolves the caller rather than leaving it parked.

`set_update_mode/2` returns `{:error, :disconnected}` rather than queueing. A queue would not survive a reboot either, and an application that needs to converge can re-assert from `c:NervesHubLink.Client.connected/0`, which is safe to repeat — the guide shows the pattern.

`device_api_version` moves to **2.4.0**, which is how NervesHub knows not to send any of this to a device that would only log it as unknown. The README's version history gains that entry and the `2.3.0` one that was never recorded.

## Docs

New [device-managed updates guide](guides/device_managed_updates.md), registered in `mix.exs` and linked from the README, with a worked example of a device that updates only inside its own maintenance window.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nshoes (approved):
🚀

